PR #4, opened 2020-06-12
Range partitioning

jfunston:
Implement manual pre-split range partitioning feature. Update TPC-C test to use it and add performance results

### DIFF
--- a/src/k2/dto/Collection.cpp
+++ b/src/k2/dto/Collection.cpp
@@ -91,7 +91,9 @@ PartitionGetter::PartitionGetter(Collection&& col) : collection(std::move(col)) 
         for (auto it = collection.partitionMap.partitions.begin();
                   it != collection.partitionMap.partitions.end();
                   ++it) {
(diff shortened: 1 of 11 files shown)


src/k2/dto/Collection.cpp
```diff
@@ -135,20 +141,27 @@ PartitionGetter::PartitionWithEndpoint& PartitionGetter::getPartitionForKey(cons
...
+    }
 }
 
 bool OwnerPartition::owns(const Key& key) const {
     switch (_scheme) {
         case HashScheme::Range:
+            if (_partition.endKey == "") {
+                return _partition.startKey.compare(key.partitionKey) <= 0;
+            }
+            // TODO should endKey comparision be strictly less than?
```
iafuture, line 164: I think so, yes. Otherwise we can't express two partitions to be [""..."b", "b"..""]. We need the key "b" to not be in the first partition in this case - it should be in the second. In other words, partitions should be [start; end)
We should also revisit hash-based partitioning to make sure it follows the same convention

PR #6, opened 2020-06-25
K23SI Correctness

jfunston:
New inspection RPCs to facilitate tests and correctness tools. Augment tests with new RPCs.

### DIFF
--- a/src/k2/dto/K23SIInspect.h
+++ b/src/k2/dto/K23SIInspect.h
@@ -1,0 +1,97 @@
+/*
+MIT License
+
+Copyright(c) 2020 Futurewei Cloud
+
+    Permission is hereby granted,
+    free of charge, to any person obtaining a copy of this software and associated documentation files(the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and / or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions :
+
+    The above copyright notice and this permission notice shall be included in all copies
+    or
+    substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS",
+    WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+    DAMAGES OR OTHER
+    LIABILITY,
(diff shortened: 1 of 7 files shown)


src/k2/dto/K23SIInspect.h
```diff
@@ -0,0 +1,97 @@
...
+struct K23SIInspectRecordsRequest {
+    // These fields make the request compatible with the PartitionRequest wrapper
+    Partition::PVID pvid; // the partition version ID. Should be coming from an up-to-date partition map
+    String collectionName;
+    Key key; // the key to gather all records for
+    K2_PAYLOAD_FIELDS(pvid, collectionName, key);
+};
+
+struct K23SIInspectRecordsResponse {
+    std::vector<DataRecord> records;
```
iafuture, line 45: I assume this is the DataRecord from 3si module?

src/k2/dto/K23SIInspect.h
```diff
@@ -0,0 +1,97 @@
...
+
+    friend std::ostream& operator<<(std::ostream& os, const K23SIInspectTxnResponse& rec) {
+        os << "{txnId=" << rec.txnId << ", writeKeys=[";
+        os << rec.writeKeys.size();
+        os << "], rwExpiry=" << rec.rwExpiry << ", syncfin=" << rec.syncFinalize << "}";
+        return os;
+    }
+};
+
+// For test and debug, requests all WIs on a node for any key
```
iafuture, line 85: is this different that inspectRecords?

src/k2/dto/K23SIInspect.h
jfunston: Yes. Are you worried about the header dependency?

src/k2/dto/K23SIInspect.h
jfunston: The comment should say "all keys". The idea is to use it for the orphaned WI scan.

src/k2/dto/K23SIInspect.h
iafuture: yes - are you moving DataRecord to DTO?

src/k2/dto/K23SIInspect.h
iafuture: so i guess we should have InspectKey (all versions for 1 key) and InspectAllKeys(all versions for all keys) ?

src/k2/dto/K23SIInspect.h
jfunston: ok, I moved it

src/k2/dto/K23SIInspect.h
jfunston: I added InspectAllKeys, but it only returns the keys themselves. I figured that returning all the versions too would be too big for one request. Sound good?

jfunston:
I added the RPC to get all TRHs. We can save the tracing support for a separate PR, I will create an issue for it.

src/k2/dto/K23SIInspect.h
iafuture: makes sense

PR #11, opened 2020-07-08
Document store

iafuture:


src/k2/cmd/demo/document_demo.cpp
```diff
@@ -0,0 +1,161 @@
...
+            writeDoc.serializeField<uint32_t, dto::DocumentFieldType::UINT32T>(777, "Balance");
+
+            return txn.write(std::move(writeDoc))
+            .then([&txn] (WriteResult&& result) {
+                K2ASSERT(result.status.is2xxOK(), "Write failed");
+                return std::move(txn);
+            });
+        })
+        .then([this] (K2TxnHandle&& txn) {
+            SerializableDocument readDoc = _client.makeSerializableDocument("demo", "demo_schema");
```
iafuture, line 117: What schema version will this use? On one hand, we want to discover records just based on schema name, not version, to allow for schema updates. On the other, we probably can't handle changes to primary/range key schema since that would mean each version needs different fields for keys...

src/k2/dto/K23SI.h
```diff
@@ -210,6 +212,41 @@ struct K23SIWriteResponse {
...
+    EQ,
+    NEQ,
+    GT,
+    GTE,
+    LT,
+    LTE,
+    STARTS_WITH
+};
+
+struct K23SIQueryRequest {
```
iafuture, line 225: it would be nice if we could add some convenience API to build predicates, something like qreq.addFilter("name", K23SIQueryOp.STARTS_WITH, "Jo");

... I just saw you added that below. ANy reason not to have the API on this struct?

src/k2/dto/SerializableDocument.h
```diff
@@ -0,0 +1,70 @@
...
+
+    template <typename FieldType, DocumentFieldType type>
+    FieldType deserializeField(const String& name);
+
+    template <typename FieldType, DocumentFieldType type>
+    FieldType deserializeField(uint32_t fieldIndex);
+
+    // We expose the payload in case the user wants to write it to file or otherwise 
+    // store it on their own. For normal K23SI operations the user does not need to touch this variable
+    Payload fieldData;
```
iafuture, line 49: I think there's an issue with exposing this directly - the cursor over it. We should instead "share" a payload when they ask for one, and rewind the cursor on it correctly to support writing it out correctly

src/k2/cmd/demo/document_demo.cpp
```diff
@@ -0,0 +1,161 @@
...
+                    dto::DocumentFieldType::STRING,
+                    dto::DocumentFieldType::STRING,
+                    dto::DocumentFieldType::UINT32T};
+            schema.fieldNames = std::vector<String>{"LastName", "FirstName", "Balance"};
+            // LastName is our partition key
+            schema.partitionKeyFields = std::vector<uint32_t>{0};
+            // FirstName is our range key
+            schema.rangeKeyFields = std::vector<uint32_t>{1};
+
+            dto::CreateSchemaRequest request{ "demo_schema", std::move(schema) };
```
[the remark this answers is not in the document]
jfunston: Yes, schemas are associated with collection. This line should have "demo" instead of "demo_schema"

PR #12, opened 2020-07-28
Schema create, get, and push implementation

jfunston:


jfunston:
A couple extra validation steps I thought of while preparing the PR: ensure there are no duplicate field names, and ensure that partitionKey and rangeKey fields do not change between schema versions. I can add these before we merge.

src/k2/cpo/service/CPOService.cpp
```diff
@@ -183,6 +195,184 @@ CPOService::handleGet(dto::CollectionGetRequest&& request) {
...
+    }
+    dto::Collection& collection = std::get<1>(getResponse);
+
+    bool validatedKeys = false;
+    for (const dto::Schema& otherSchema : schemas[request.collectionName]) {
+        if (otherSchema.name == request.schema.name && otherSchema.version == request.schema.version) {
+            return RPCResponse(Statuses::S403_Forbidden("Schema name and version already exist"), dto::CreateSchemaResponse{});
+        }
+
+        // By induction, as long as we validate that the key fields match with at least one
```
iafuture, line 345: I think it's a bit confusing to say that this is true by induction. i associate this with recursion, and i'd expect to see comparison with the latest version to have a proof by induction. I think i can convince myself that compatibility with any version implies compatibility of the entire chain, but in my head, that's because the only mutating operations allowed are add/rename/drop column[s], and all of these are idempotent. so maybe just a comment thing :)

src/k2/cpo/service/CPOService.cpp
```diff
@@ -183,6 +195,184 @@ CPOService::handleGet(dto::CollectionGetRequest&& request) {
...
+        }
+    }
+
+
+    // 3. Update in memory
+
+    schemas[request.collectionName].push_back(std::move(request.schema));
+
+    // 4. Save to disk
+    Status saved = _saveSchemas(request.collectionName);
```
iafuture, line 363: if save to disk can fail, it maybe better to attempt to save first before updating in-memory state

src/k2/cpo/service/CPOService.cpp
```diff
@@ -268,9 +458,36 @@ std::tuple<Status, dto::Collection> CPOService::_getCollection(String name) {
...
+            std::get<0>(result) = Statuses::S500_Internal_Server_Error("unable to read schema data");
+            return result;
+        }
+    }
+
     return result;
 }
 
+Status CPOService::_loadSchemas(const String& collectionName) {
+    auto cpath = _getCollectionPath(collectionName) + ".schemas";
```
iafuture, line 476: i see the ".schemas" append in two places. You should extract that into a helper method _getSchemasPath(collName)

src/k2/dto/ControlPlaneOracle.h
```diff
@@ -60,5 +63,76 @@ struct CollectionGetResponse {
     K2_PAYLOAD_FIELDS(collection);
 };
 
+struct SchemaField {
+    DocumentFieldType type;
+    String name;
+    // Ascending or descending sort order. Currently only relevant for 
+    // key fields, but could be used for secondary index in the future
+    bool descending;
```
iafuture, line 71: you should set defaults for these fields (e.g. bool descending=false;)

src/k2/dto/ControlPlaneOracle.h
```diff
@@ -60,5 +63,76 @@ struct CollectionGetResponse {
...
+    // NULL first or last in sort order. Relevant for key fields and 
+    // for open-ended filter predicates
+    bool nullLast; 
+    K2_PAYLOAD_FIELDS(type, name, descending, nullLast);
+};
+
+struct Schema {
+    String name;
+    uint32_t version;
```
iafuture, line 80: default value

src/k2/dto/K23SI.h
```diff
@@ -357,5 +358,15 @@ struct K23SITxnFinalizeRequest {
 struct K23SITxnFinalizeResponse {
     K2_PAYLOAD_EMPTY;
 };
+
+struct K23SIPushSchemaRequest {
+    Schema schema;
```
iafuture, line 363: i think the way you have Schema now, we should also specify which collection this is for. In other words, the global unique key for a schema is
<collection_id>:<schema_id>:<schema_version> (with schema_id<->schema_name being 1<->1).  Else, we need the collection name in the Schema object itself, which considering how things are structured, might be a fine idea too.

src/k2/module/k23si/Module.cpp
```diff
@@ -561,6 +566,15 @@ K23SIPartitionModule::handleTxnFinalize(dto::K23SITxnFinalizeRequest&& request)
     });
 }
 
+seastar::future<std::tuple<Status, dto::K23SIPushSchemaResponse>>
+K23SIPartitionModule::handlePushSchema(dto::K23SIPushSchemaRequest&& request) {
+    K2DEBUG("handlePushSchema for schema: " << request.schema.name);
+
```
iafuture, line 572: this should validate to make sure this schema is for the module collection

src/k2/cpo/service/CPOService.cpp
```diff
@@ -183,6 +195,184 @@ CPOService::handleGet(dto::CollectionGetRequest&& request) {
...
+        K2ERROR_EXC("Failed to push schema update: ", exc);
+        return seastar::make_ready_future<Status>(Statuses::S500_Internal_Server_Error("Failed to push schema"));
+    });
+}
+
+seastar::future<std::tuple<Status, dto::GetSchemasResponse>>
+CPOService::handleSchemasGet(dto::GetSchemasRequest&& request) {
+    auto it = schemas.find(request.collectionName);
+    if (it == schemas.end()) {
+        std::tuple<Status, dto::Collection> getResponse = _getCollection(request.collectionName);
```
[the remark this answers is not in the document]
jfunston: I need the partition map from the collection for the schema push, and it also serves as a cache refresh.

src/k2/cpo/service/CPOService.cpp
```diff
@@ -183,6 +195,184 @@ CPOService::handleGet(dto::CollectionGetRequest&& request) {
...
+        if (!std::get<0>(getResponse).is2xxOK()) {
+            return RPCResponse(std::move(std::get<0>(getResponse)), dto::GetSchemasResponse{});
+        }
+
+        it = schemas.find(request.collectionName);
+        K2ASSERT(it != schemas.end(), "Schemas iterator is end after collection refresh");
```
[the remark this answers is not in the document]
jfunston: No defined schemas is a valid case, but for that we should find it in the map but the vector will be empty. That is how I tell the difference between a collection with no schemas and the case where the schemas need to be read from disk.

PR #13, opened 2020-08-04
SerializableKV Record implementation and basic unit tests

jfunston:


src/k2/dto/SKVRecord.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+SKVRecord::SKVRecord(const String& collection, Schema s) : 
+            collectionName(collection), schemaName(s.name), schemaVersion(s.version), schema(s) {
+    fieldData = Payload([] () { return Binary(DEFAULT_SEGMENT_SIZE); });
+    partitionKeys.resize(schema.partitionKeyFields.size());
+    rangeKeys.resize(schema.partitionKeyFields.size());
+}
+
+String SKVRecord::getPartitionKey() {
+    String partitionKey("");
```
iafuture, line 101: Is there some way to construct this in more optimal way - in one shot so that we don't have to re-allocate the string for every append below

src/k2/dto/SKVRecord.cpp
```diff
@@ -0,0 +1,125 @@
...
+
+namespace k2 {
+namespace dto {
+
+void SKVRecord::skipNext() {
+    if (fieldCursor >= schema.fields.size()) {
+        throw new std::runtime_error("Schema not followed in record serialization");
+    }
+
+    for (size_t i = 0; i < schema.partitionKeyFields.size(); ++i) {
```
iafuture, line 37: this seems like a problem for range keys

src/k2/dto/SKVRecord.h
```diff
@@ -0,0 +1,172 @@
...
+public:
+    // The record must be serialized in order. Schema will be enforced
+    template <typename T>
+    void serializeNext(T field) {
+        FieldType ft = TToFieldType<T>();
+        if (fieldCursor >= schema.fields.size() || ft != schema.fields[fieldCursor].type) {
+            throw new std::runtime_error("Schema not followed in record serialization");
+        }
+
+        for (size_t i = 0; i < schema.partitionKeyFields.size(); ++i) {
```
iafuture, line 44: This seems like a problem for range keys

src/k2/dto/SKVRecord.h
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        throw new std::runtime_error("Schema not followed in record deserialization");
+    }
+
+    void seekField(uint32_t fieldIndex);
+
+    template <typename T>
+    T deserializeField(uint32_t fieldIndex) {
```
iafuture, line 76: How does this deal with optional fields? I'd expect same response as the deserializeNextOptional() method

PR #16, opened 2020-08-14
update TXN spec

jerryhfeng:
Update the TXN.md with retentionWindow and snapshot-backed read related info, and rename it to K2-3SI-TXN.md


docs/K2-3SI-TXN.md
```diff
@@ -160,6 +162,8 @@ To remain correct, we allow the users to specify the data retention window. Node
 1. Any transaction commit with `ts < time.NOW - retentionWindow` will be aborted
 1. Any in-progress transactions running longer than the retentionWindow will be aborted (`ForcedAbort`)
 
+Normally a transaction has to start and finish within the retention window. It is usually not an issue for update kind of OLTP transactions which is typically short running. For for long running OLAP read-only transactions, such retention window may not provide long enoug time. For such long running read-only transaction, we will provide snapshot-backed-read feature, which system will transparently take and keep the snapshot version of the data for the transaction to read, and once the read-only transaction completes, such transparent snapshot will be deleted with the data it holds if they are out of retention window. Of course, even for such snapshot-backed read transaction, there is a max time-out value to prevent it holds the snapshot too long. Ultimately, if such transparent snapshot-backed read is not sufficient, the user could take a snapshot (and bear the cost of it) and issue the read-only transaction on the real user snapshot.
```
iafuture, line 165: I'm a bit unclear on this proposal. It seems to imply that the users could at some point instruct the cluster to temporarily hold onto a snapshot of their data, and that they would enjoy the same level of consistency as usual.
I think we should talk about providing read-committed level of isolation with read-your writes consistency. To do so, we do not have to lock the data in the cluster to support the transaction, we simply need to return the latest committed data after the transaction timestamp.

jerryhfeng: Indeed we could provide ReadCommitted isolation read as well, as an RFC already pushed in. But here, a snapshot isolation read is proposed. With such auto snapshot backed read feature, we could reduce the data retention windows size, i.e. lockdown less version of data.

jerryhfeng: There are various reason such auto snapshot-backed read transaction to be failed, e.g. read-write conflict, and more tricky one like the read come to a partition for the first time inside the transaction when it is too late, beyond the retention window, etc. It is hard to explain and educate user for these cases. For now, let's skip such auto snapshot-backed read transaction.

jerryhfeng:
Will update.

PR #18, opened 2020-08-14
Key Encoding RFC and escaped key string implementation

jfunston:


src/k2/dto/FieldTypes.cpp
```diff
@@ -33,51 +33,97 @@ Copyright(c) 2020 Futurewei Cloud
...
 
 template <> FieldType TToFieldType<String>() { return FieldType::STRING; }
 template <> FieldType TToFieldType<uint32_t>() { return FieldType::UINT32T; }
 
 // All conversion assume ascending ordering
 
-template <> String FieldToKeyString<String>(const String& field) {
-    // Sizes will not match if there are exta null bytes
-    K2ASSERT(field.size() == strlen(field.c_str()), "String has null bytes");
+static String escapeKeyString(const String& field) {
```
iafuture, line 45: For efficiency, i wonder if it also makes sense to provide a concat & escape method, where you can perform the escaping and terminator-based concatenation in one allocation.

src/k2/dto/FieldTypes.cpp
```diff
@@ -33,51 +33,97 @@ Copyright(c) 2020 Futurewei Cloud
...
+    std::vector<size_t> foundNulls;
+
+    while (pos != String::npos && pos < field.size()) {
+        pos = field.find(ESCAPE, pos);
+        if (pos != String::npos) {
+            foundNulls.push_back(pos);
+            ++pos;
+        }
+    }
+
```
iafuture, line 56: maybe shortcut the function here: if no nulls were found just return the original

src/k2/dto/FieldTypes.cpp
iafuture: Also, i think it'd be better if this method is a noop by itself if the string doesn't need escaping. That way you don't have to check for noop before calling it like you do below. I think one way to do this is via a signature
bool escape(const String& in, String& out). If no-op, we don't set the output.
Another way is to use
String escape(String&& in). If no-op, we set the output to the input by move.

docs/RFC/key_encoding.md
```diff
@@ -0,0 +1,153 @@
...
+Then each of the encoded fields in the key are simply concatenated in the order specifiec by the schema. 
+This fulfills all five of our desired properties.
+
+
+### Implementation Notes
+
+The client library is responsible for encoding the partition and range keys from the record's field values. 
+It must do this for the partition key so that the request can be routed, so for simplicity it will also 
+do this for the range key. When the server gets a request it concatenates SCHEMA\_NAME + TERMINATOR + 
+PARTITION\_KEY + RANGE\_KEY for indexer operations. The server does not need to do further encoding beyond
```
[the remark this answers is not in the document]
jfunston: The last fields of the partition key and range key will already have the terminator, so the server doesn't need to add them.

src/k2/dto/FieldTypes.cpp
jfunston: I combined them to one function to avoid the copy and make it a bit cleaner.

PR #19, opened 2020-08-20
Modified indexer to use templated class

### LINKED ISSUE #29: HOT integration

ashayshirwadkar:
Currently only map indexer is implemented as a specialized class
of the templated class.

Fixes #29


src/k2/module/k23si/Module.cpp
```diff
@@ -697,6 +702,7 @@ K23SIPartitionModule::handleInspectAllKeys(dto::K23SIInspectAllKeysRequest&& req
     std::vector<dto::Key> keys;
     keys.reserve(_indexer.size());
 
+
```
ashayshirwadkar, line 705: Remove space

ashayshirwadkar: Done

src/k2/indexer/IndexerInterface.h
```diff
@@ -22,18 +22,63 @@ Copyright(c) 2020 Futurewei Cloud
...
 //
 //  K2 internal MVCC representation
 //
-template<typename ValueType>
-struct VersionedTreeNode {
-    uint64_t version;
-    ValueType value;
-    std::unique_ptr<VersionedTreeNode<ValueType>> next;
+
+template<typename IndexerType, typename ValueType>
```
iafuture, line 38: This sort of template args would require a bunch of redundant type specifications which may also lead to some compile issues at best, or runtime bugs at worst.
Could we use the map as template argument instead?
e.g. usage:
Indexer<std::map, dto::Key, dto::Record> idxr;
idxr.find(dto::Key{"a","b"});

PR #21, opened 2020-08-24
Added the Plog support

### LINKED ISSUE #23: Add plog support
References:
src/k2/dto/Persistence.h (Dtos for the communication between PlogServer and PlogClient)
src/k2/dto/PartitionGroup.h (Dtos for register/get plog partition groups)
src/k2/persistence/plog/PlogServer.h
src/k2/persistence/plog/PlogServer.cpp
src/k2/persistence/plog/PlogClient.h
src/k2/persistence/plog/PlogClient.cpp
src/k2/cmd/demo/plog_server.cpp (Start the PlogServer instance)
test/plog/* 
test/integration/test_plog.sh (Intergration unit test cases for plog service)

piggesthjy:
Added the Plog support. fixes #23


src/k2/cmd/demo/plog_server.cpp
```diff
@@ -0,0 +1,33 @@
+/*
```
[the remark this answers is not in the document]
piggesthjy: Rename the plog_server.cpp to plog_main and put it into the cmd/plog folder

src/k2/cpo/client/CPOClient.h
```diff
@@ -233,6 +234,44 @@ class CPOClient {
...
+                status = Statuses::S408_Request_Timeout("partition group deadline exceeded");
+                return seastar::make_ready_future<Status>(std::move(status));
+            }
+            return seastar::make_ready_future<Status>(std::move(status));
+        });
+    }
+
+    template<typename ClockT=Clock>
+    seastar::future<std::tuple<Status, dto::PartitionMapGetResponse>> GetPartitionMap(Deadline<ClockT> deadline) {
+        dto::PartitionMapGetRequest request{.offset = 0};
```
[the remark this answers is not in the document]
piggesthjy: Fixed this part by changing the GetPartitionMap API to GetPartitionCluster API

src/k2/cmd/demo/plog_server.cpp
```diff
@@ -0,0 +1,33 @@
+/*
+MIT License
+
```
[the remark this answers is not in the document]
piggesthjy: Since the persistentVolume requires the older persistence service, I created another 2 folders in persistence folder: plog_client and plog_server

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    K2DEBUG("Received read request for " << request.plogId << " with offset " << request.offset);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S400_Bad_Request("plog does not exist"), dto::PlogReadResponse());
+    }
+    if (iter->second.offset < request.offset + request.size){
+         return RPCResponse(Statuses::S400_Bad_Request("exceed the maximun length"), dto::PlogReadResponse());
+    }
+    
+    Payload payload([size=request.size] { return Binary(size); });
```
iafuture, line 128: reads shouldn't require a copy.I think we can enhance Payload to provide a method to share data into another payload:
```c++
Payload Payload::share(size_t nbytes) 
```

src/k2/persistence/plog_client/PlogClient.h
```diff
@@ -0,0 +1,77 @@
...
+private:
+    dto::PartitionCluster _partitionCluster;
+    std::unordered_map<String, std::vector<std::unique_ptr<TXEndpoint>>> _partitionMapEndpoints;
+    std::vector<String> _partitionNameList;
+    uint32_t _partition_map_pointer;
+
+    CPOClient _cpo;
+    String generate_plogId();
+
+    ConfigDuration get_plog_server_deadline{"get_plog_server_deadline", 1s};
```
iafuture, line 71: use underscore prefix for the private variable names. Also call this _cpo_timeout instead since i think it represents better what it is supposed to do

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+#include <cstdlib>
+#include <cctype>
+#include <algorithm>
+
+namespace k2 {
+
+char Rand_alnum()
+{
+    char c;
+    while (!std::isdigit(c = static_cast<char>(std::rand())));
```
iafuture, line 44: this seems an expensive way to generate a char that is a number. how about this:
``` c++

return "0123456789"[std::rand()%10];
```
even better if you do random without using modulo and a faster rand generator, such as http://prng.di.unimi.it/

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    return seastar::make_ready_future<>();
+}
+
+seastar::future<std::tuple<Status, dto::PlogCreateResponse>>
+PlogServer::handleCreate(dto::PlogCreateRequest&& request){
+    K2DEBUG("Received create request for " << request.plogId);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter != _plogMap.end()) {
+        return RPCResponse(Statuses::S400_Bad_Request("plog already exists"), dto::PlogCreateResponse());
```
iafuture, line 82: better code needed here. You can look at the standard status codes in Statuses::

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+                sealed_offset = response.offset;
+                if (!return_status.is2xxOK()) 
+                    break;
+            }
+            return seastar::make_ready_future<std::tuple<Status, uint32_t> >(std::tuple<Status, uint32_t>(std::move(return_status), std::move(sealed_offset)));
+        });
+}
+
+String PlogClient::generate_plogId(){
+    std::string appendix;
```
iafuture, line 176: I think it'd be more efficient if you start with a string that is big enough to hold the prefix and the numeric code and then just shuffle the last chars
https://en.cppreference.com/w/cpp/algorithm/random_shuffle
``` c++
    String plogid = "TPCC_CLIENT_plog_0123456789";
    std::mt19937 g(rd());
    std::shuffle(plogid.begin()+18, plogid.end(), g);
```

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+PlogClient::PlogClient() {
+    K2INFO("dtor");
+}
+
+PlogClient::~PlogClient() {
+    K2INFO("~dtor");
+}
+
+void PlogClient::GetPlogServerEndpoint() {
```
iafuture, line 56: not sure why this method starts with upper case. Also, it should be private and I think its purpose is to fetch the plogPartitionMap, not the plogServerEndpoint - it is populating the entire partition map. You should change the name i think to getPlogPartitionMap

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+    for(auto& v : _partitionCluster.partitionGroupVector){
+        K2INFO("Partition Group: " << v.name);
+        _partitionNameList.push_back(v.name);
+        
+        std::vector<std::unique_ptr<TXEndpoint>> endpoints;
+        for (auto& url: v.plogServerEndpoints){
+            K2INFO("Plog Server Url: " << url);
+            auto ep = RPC().getTXEndpoint(url);
```
iafuture, line 64: this could be empty if url is bad. don't push it in if it is empty

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+            endpoints.push_back(std::move(ep));
+        }
+
+        _partitionMapEndpoints[std::move(v.name)] = std::move(endpoints);
```
iafuture, line 68: same here - this vector may be empty. in which case you should return an exceptional future

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+
+        if (!status.is2xxOK()) {
+            K2INFO("Failed to obtain Partition Cluster" << status);
+            return seastar::make_exception_future<>(std::runtime_error("Failed to obtain Partition Cluster Map"));
+        }
+
+        _partitionCluster = std::move(response.cluster);
+        _partition_map_pointer = rand() % _partitionCluster.partitionGroupVector.size();
+        _partitionMapEndpoints.clear();
+        GetPlogServerEndpoint();
```
iafuture, line 87: this method should return a future based on whether it succeeded or not. Then here you'd just: `return getPlogServerEndpoint()`

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+        return seastar::make_ready_future<>();
+    });
+}
+
+seastar::future<std::tuple<Status, String>> PlogClient::create(){
```
iafuture, line 92: your plog creation generates a random id. It is possible to fail due ID collision. You should retry in that case with a new ID.

Also, if this fails, we should try and create the plog in another partition.

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,182 @@
...
+    then([this] (auto&& result) {
+        auto& [status, response] = result;
+
+        if (!status.is2xxOK()) {
+            K2INFO("Failed to obtain Partition Cluster" << status);
+            return seastar::make_exception_future<>(std::runtime_error("Failed to obtain Partition Cluster Map"));
+        }
+
+        _partitionCluster = std::move(response.cluster);
+        _partition_map_pointer = rand() % _partitionCluster.partitionGroupVector.size();
```
iafuture, line 85: i'm not sure if random choice is the best for selecting the partition group. I think we should allow this to be set by the user of the library

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+
+seastar::future<std::tuple<Status, dto::PlogSealResponse>>
+PlogServer::handleSeal(dto::PlogSealRequest&& request){
+    K2DEBUG("Received seal request for " << request.plogId);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S400_Bad_Request("plog does not exist"), dto::PlogSealResponse());
+    }
+    if (iter->second.sealed){
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    return RPCResponse(Statuses::S200_OK("read success"), std::move(response));
+};
+
+
+seastar::future<std::tuple<Status, dto::PlogSealResponse>>
+PlogServer::handleSeal(dto::PlogSealRequest&& request){
+    K2DEBUG("Received seal request for " << request.plogId);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S400_Bad_Request("plog does not exist"), dto::PlogSealResponse());
```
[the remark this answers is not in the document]
piggesthjy: fixed

test/transport/PayloadTest.cpp
```diff
@@ -276,6 +289,33 @@ SCENARIO("test empty payload serialization after some data") {
         REQUIRE(dst2.copy() == dst2);
     }
 }
+
+SCENARIO("test cooy from payload") {
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.h
```diff
@@ -0,0 +1,83 @@
...
+
+    bool sealed;
+    uint32_t offset;
+    Payload payload;
+};
+
+class PlogServer
+{
+private:
+    uint32_t PLOG_MAX_SIZE = 2 * 1024 * 1024;
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    _plogMap.insert(std::pair<String,PlogPage >(std::move(request.plogId), PlogPage()));
+    return RPCResponse(Statuses::S201_Created("plog created"), dto::PlogCreateResponse());
+};
+
+seastar::future<std::tuple<Status, dto::PlogAppendResponse>>
+PlogServer::handleAppend(dto::PlogAppendRequest&& request){
+    K2DEBUG("Received append request for " << request.plogId << " with size" << request.payload.getSize() << " and offset " << request.offset);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S400_Bad_Request("plog does not exist"), dto::PlogAppendResponse());
```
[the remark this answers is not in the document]
piggesthjy: I have changed all the status to be a suitable one

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+
+    dto::PlogAppendResponse response;
+    response.offset = iter->second.offset + request.payload.getSize();
+    response.bytes_appended = request.payload.getSize();
+
+    iter->second.offset += request.payload.getSize();
+    iter->second.payload.copyFromPayload(request.payload, request.payload.getSize());
+    request.payload.clear();
+    
+    return RPCResponse(Statuses::S200_OK("append cuccess"), std::move(response));
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+         return RPCResponse(Statuses::S400_Bad_Request("exceeds pLog limit"), dto::PlogAppendResponse());
+    }
+
+    dto::PlogAppendResponse response;
+    response.offset = iter->second.offset + request.payload.getSize();
+    response.bytes_appended = request.payload.getSize();
+
+    iter->second.offset += request.payload.getSize();
+    iter->second.payload.copyFromPayload(request.payload, request.payload.getSize());
+    request.payload.clear();
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: I added the Payload Payload::share(size_t nbytes) in Payload.h and fixed this problem

src/k2/persistence/plog_client/PlogClient.h
piggesthjy: fixed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: I used the shuffle method to generate the plog id so that we do not need this one

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: fixed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: fixed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: fixed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: I throw an exexception when the vector is empty

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,169 @@
...
+    dto::PlogReadResponse response;
+    response.payload = std::move(payload);
+    return RPCResponse(Statuses::S200_OK("read success"), std::move(response));
+};
+
+
+seastar::future<std::tuple<Status, dto::PlogSealResponse>>
+PlogServer::handleSeal(dto::PlogSealRequest&& request){
+    K2DEBUG("Received seal request for " << request.plogId);
+    auto iter = _plogMap.find(request.plogId);
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: fixed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: I added an API `bool selectPartitionGroup(String name);` so that the user can select the partition group by the name of the partition group

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: added the retry for ID collision and list the retry for other cases as TODO

src/k2/transport/Status.h
```diff
@@ -450,6 +450,11 @@ static const inline Status S527_Railgun_Error{.code=527, .message=""};
 // Used by Qualys in the SSLLabs server testing API to signal that the site can't process the request.
 static const inline Status S529_Site_is_overloaded{.code=529, .message=""};
 
+// 530 Plog Id is duplicated 
+// Used by Plog Server to signal that the Plog Id is duplicated
+static const inline Status S530_PlogId_is_duplicated{.code=530, .message=""};
```
iafuture, line 455: You should not create custom codes. This file contains all HTTP standard status codes. We should be able to find one which suits . In particular, this error is not retryable so it should not be 5xx series code but rather 4xx series. I think 409_Conflict would work well here. A bit less descriptive but also usable code is 403_Forbidden

src/k2/transport/Payload.h
```diff
@@ -120,6 +120,9 @@ class Payload {
     // the underlying data will not be destroyed until all shared payloads are destroyed.
     Payload share();
 
+    // Same as shart(), but just share nbytes
```
iafuture, line 123: did you mean "share" ?

src/k2/transport/Payload.cpp
```diff
@@ -391,6 +391,34 @@ Payload Payload::share() {
     return shared;
 }
 
+
+Payload Payload::share(size_t nbytes) {
```
iafuture, line 395: This code is nearly identical to share() but it is different. share() shares the entire payload, regarding of current position, so share(nbytes) doesn't quite make sense as a method name and intended functionality.
I would rework the api a bit to reflect that:
Payload shareAll() ; // share entire payload
Payload shareRegion(PayloadPosition start, size_t nbytes); // share a particular region from the payload.

the new `shareRegion` method would be a slightly modified version of the existing `share()` method.

then shareAll() implementation is just:
```c++
return shareRegion(PayloadPosition(0,0,0), getSize());
```

test/transport/PayloadTest.cpp
```diff
@@ -276,6 +289,33 @@ SCENARIO("test empty payload serialization after some data") {
         REQUIRE(dst2.copy() == dst2);
     }
 }
+
```
iafuture, line 292: can you please add a test for `shareRegion()` and `shareAll()`?

test/plog/PlogTest.cpp
```diff
@@ -0,0 +1,224 @@
...
+}
+
+seastar::future<> PlogTest::runTest3() {
+    K2INFO(">>> Test3: read the persistence group we created in test2");
+    return _client.getPersistenceCluster("Cluster1")
+    .then([this] () {
+        K2INFO("Test3.1: create a plog");
+        return _client.create();
+    })
+    .then([this] (auto&& response){
```
iafuture, line 113: you should add a test for double-creation (server should reject a create for existing plog)

src/k2/persistence/plog_service/PlogServer.h
```diff
@@ -0,0 +1,83 @@
...
+#include <k2/dto/Plog.h>
+#include <k2/common/Common.h>
+#include <k2/config/Config.h>
+#include <k2/cpo/client/CPOClient.h>
+#include <k2/appbase/AppEssentials.h>
+#include <k2/appbase/Appbase.h>
+#include <k2/transport/BaseTypes.h>
+#include <k2/transport/TXEndpoint.h>
+
+namespace k2
```
iafuture, line 39: please add comments in the file

src/k2/persistence/plog_client/PlogClient.h
```diff
@@ -0,0 +1,81 @@
...
+class PlogClient {
+public:
+    PlogClient();
+    ~PlogClient();
+
+    seastar::future<> gracefulStop();
+
+    seastar::future<> start();
+
+    seastar::future<> getPlogPersistenceMap();
```
iafuture, line 51: i don't understand these getters and select. Why are these getters public, and why do they not return anything?
I think the intent here is to allow users to obtain the cluster info so that they can select a partition. In that case, you should have a method
```
const dto::PersistenceCluster& getCluster();
```
which returns the cluster for which this plog client is created.

src/k2/persistence/plog_client/PlogClient.h
```diff
@@ -0,0 +1,81 @@
...
+#include <k2/cpo/client/CPOClient.h>
+#include <k2/transport/BaseTypes.h>
+#include <k2/transport/TXEndpoint.h>
+#include <k2/appbase/AppEssentials.h>
+#include <k2/appbase/Appbase.h>
+
+
+namespace k2 {
+
+class PlogClient {
```
iafuture, line 42: I don't quite understand how this plog client is supposed to be used due to the lack of comments and what I think is a bit confusing API.  You should add enough comments to explain, and if you have an example usage in another file, you should reference it here.

I am not sure if this was the intent here, but  I would expect a client to be created with some configuration so that it can initialize correctly. An important piece of configuration is the cluster. I think requiring 1 client per cluster is a good idea, in which case the cluster name is something the user should be required to provide when creating a plog client.

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+    return seastar::make_ready_future<>();
+}
+
+seastar::future<std::tuple<Status, dto::PlogCreateResponse>>
+PlogServer::_handleCreate(dto::PlogCreateRequest&& request){
+    K2DEBUG("Received create request for " << request.plogId);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter != _plogMap.end()) {
+        return RPCResponse(Statuses::S530_PlogId_is_duplicated("plog id already exists"), dto::PlogCreateResponse());
```
iafuture, line 82: don't create your own code. Use an existing one with custom message

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,161 @@
...
+PlogServer::_handleSeal(dto::PlogSealRequest&& request){
+    K2DEBUG("Received seal request for " << request.plogId);
+    dto::PlogSealResponse response;
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S404_Not_Found("plog does not exist"), std::move(response));
+    }
+    if (iter->second.sealed){
+        response.offset = iter->second.offset;
+        return RPCResponse(Statuses::S409_Conflict("plog already sealed"), std::move(response));
```
iafuture, line 143: I think sealing should be an immutable operation - if a plog has been sealed, it should be ok to seal again

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+
+seastar::future<std::tuple<Status, dto::PlogReadResponse>>
+PlogServer::_handleRead(dto::PlogReadRequest&& request){
+    K2DEBUG("Received read request for " << request.plogId << " with offset " << request.offset);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S404_Not_Found("plog does not exist"), dto::PlogReadResponse());
+    }
+    if (iter->second.offset < request.offset + request.size){
```
iafuture, line 123: you should use 403_forbidden here. 400_bad_request is a general "something is bad with your request" code. It could also be that they passed "hello" as the offset. 403_forbidden is more specific and is used when the request is fine syntactically but semantically the service isn't accepting it.

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,161 @@
...
+         return RPCResponse(Statuses::S400_Bad_Request("exceed the maximun length"), dto::PlogReadResponse());
+    }
+    
+    iter->second.payload.seek(request.offset);
```
iafuture, line 127: I think with my proposed change to implement a `shareRegion` api in the payload, you don't have to seek here. For completeness though, i wanted to point out that you have to remember the current payload offset, and then seek to it after you are done sharing so that the payload is ready to append new data.

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,161 @@
...
+    K2DEBUG("Received append request for " << request.plogId << " with size" << request.payload.getSize() << " and offset " << request.offset);
+    auto iter = _plogMap.find(request.plogId);
+    if (iter == _plogMap.end()) {
+        return RPCResponse(Statuses::S404_Not_Found("plog does not exist"), dto::PlogAppendResponse());
+    }
+    if (iter->second.sealed){
+         return RPCResponse(Statuses::S409_Conflict("plog is sealed"), dto::PlogAppendResponse());
+    }
+    if (iter->second.offset != request.offset){
+        return RPCResponse(Statuses::S400_Bad_Request("offset inconsistent"), dto::PlogAppendResponse());
```
iafuture, line 99: you should try to return different codes if you want the user to handle them differently. you return 400_bad_request in both cases here.

In particular, a mismatch in offset i think is a good candidate for 403_Forbidden, while 413 Request Entity Too Large i think works better for the case where the user wants to append too much data

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -0,0 +1,161 @@
...
+    if (iter->second.offset + request.payload.getSize() > PLOG_MAX_SIZE){
+         return RPCResponse(Statuses::S400_Bad_Request("exceeds pLog limit"), dto::PlogAppendResponse());
+    }
+
+    dto::PlogAppendResponse response;
+    response.offset = iter->second.offset + request.payload.getSize();
+    response.bytes_appended = request.payload.getSize();
+
+    iter->second.offset += request.payload.getSize();
+    iter->second.payload.copyFromPayload(request.payload, request.payload.getSize());
```
iafuture, line 110: add a comment here to mention that we really want to copy in order to prevent memory fragmentation if we were to use share() instead

src/k2/cpo/service/CPOService.cpp
```diff
@@ -302,6 +311,10 @@ String CPOService::_getCollectionPath(String name) {
     return _dataDir() + "/" + name + ".collection";
 }
 
+String CPOService::_getPersistenceClusterMapPath() {
```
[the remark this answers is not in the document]
piggesthjy: Renamed it to _getPersistenceClusterPath and take a parameter

src/k2/cpo/service/CPOService.cpp
```diff
@@ -441,4 +454,56 @@ Status CPOService::_saveSchemas(const String& collectionName) {
     return Statuses::S201_Created("schema written");
 }
 
+seastar::future<std::tuple<Status, dto::PersistenceClusterCreateResponse>>
+CPOService::handlePersistenceClusterCreate(dto::PersistenceClusterCreateRequest&& request){
+    K2INFO("Received persistence cluster create request for " << request.cluster.name);
+    auto cpath = _getPersistenceClusterMapPath();
+    std::unordered_map<String, dto::PersistenceCluster> persistenceClusterMap;
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/cpo/service/CPOService.cpp
```diff
@@ -441,4 +454,56 @@ Status CPOService::_saveSchemas(const String& collectionName) {
...
+        return RPCResponse(Statuses::S500_Internal_Server_Error("unable to write persistence cluster map data"), dto::PersistenceClusterCreateResponse());
+    }
+    return RPCResponse(Statuses::S201_Created("persistence cluster creates successfully"), dto::PersistenceClusterCreateResponse());
+}
+
+seastar::future<std::tuple<Status, dto::PersistenceClusterGetResponse>>
+CPOService::handlePersistenceClusterGet(dto::PersistenceClusterGetRequest&& request) {
+    K2INFO("Received persistence cluster get request with name " << request.name);
+    auto cpath = _getPersistenceClusterMapPath();
+    std::unordered_map<String, dto::PersistenceCluster> persistenceClusterMap;
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/dto/MessageVerbs.h
```diff
@@ -60,6 +64,10 @@ enum Verbs : k2::Verb {
 
     /************ K23SI Persistence *****************/
     K23SI_Persist = 40,
+    K23SI_PERSISTENT_CREATE,
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/dto/PersistenceGroup.h
```diff
@@ -0,0 +1,67 @@
+/*
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/dto/Plog.h
```diff
@@ -0,0 +1,98 @@
+/*
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/dto/Plog.h
```diff
@@ -0,0 +1,98 @@
...
+
+struct PlogAppendRequest {
+    String plogId;
+    uint32_t offset;
+    Payload payload;
+    K2_PAYLOAD_FIELDS(plogId, offset, payload);
+};
+
+struct PlogAppendResponse {
+    uint32_t offset;
```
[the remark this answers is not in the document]
piggesthjy: changed it to newOffset

src/k2/dto/Plog.h
```diff
@@ -0,0 +1,98 @@
...
+};
+
+struct PlogReadResponse {
+    Payload payload;
+    K2_PAYLOAD_FIELDS(payload);
+};
+
+struct PlogSealRequest {
+    String plogId;
+    uint32_t offset;
```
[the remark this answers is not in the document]
piggesthjy: changed it to truncateOffset

src/k2/dto/Plog.h
```diff
@@ -0,0 +1,98 @@
...
+    K2_PAYLOAD_FIELDS(plogId, offset);
+};
+
+struct PlogSealResponse {
+    uint32_t offset;
```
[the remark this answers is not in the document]
piggesthjy: change it to sealedOffset

src/k2/dto/Plog.h
```diff
@@ -0,0 +1,98 @@
...
+    K2_PAYLOAD_FIELDS(offset);
+};
+
+struct PlogServerRegisterRequest{
```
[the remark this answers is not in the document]
piggesthjy: removed

src/k2/dto/Plog.h
```diff
@@ -0,0 +1,98 @@
...
+    String endpoint;
+    K2_PAYLOAD_FIELDS(endpoint);
+};
+
+struct PlogServerRegisterResponse{
+    K2_PAYLOAD_EMPTY;
+};
+
+struct PlogServerGetRequest{
```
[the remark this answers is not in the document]
piggesthjy: removed

src/k2/transport/Status.h
piggesthjy: fixed

src/k2/transport/Payload.h
piggesthjy: fixed

src/k2/transport/Payload.cpp
piggesthjy: rewrite the share to shareAll and shareRegion

src/k2/persistence/plog_client/PlogClient.h
piggesthjy: rename this function to _getPlogServerEndpoints so that it will make more sense

src/k2/persistence/plog_client/PlogClient.h
piggesthjy: added an init function

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: added the comments

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: fixed

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: put the seek inside the shareRegion

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,203 @@
...
+                if (!return_status.is2xxOK()) 
+                    break;
+            }
+            return seastar::make_ready_future<std::tuple<Status, uint32_t> >(std::tuple<Status, uint32_t>(std::move(return_status), std::move(sealed_offset)));
+        });
+}
+
+String PlogClient::_generatePlogId(){
+    String plogid = "TPCC_CLIENT_plog_0123456789";
+    std::mt19937 g(std::rand());
```
iafuture, line 188: I think it may be faster if you keep this as a member variable instead of creating a new one each time you want to generate an id

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,203 @@
...
+    return seastar::when_all_succeed(createFutures.begin(), createFutures.end())
+        .then([this, plogId](std::vector<std::tuple<Status, dto::PlogCreateResponse> >&& results) { 
+            Status return_status;
+            for (auto& result: results){
+                auto& [status, response] = result;
+                return_status = std::move(status);
+                if (!return_status.is2xxOK()) 
+                    break;
+            }
+            if (return_status.code == 530){
```
iafuture, line 117: I think this code is now 409 conflict.
Also, you have an infinite loop possibility (e.g. if the server has a bug and keeps returning 409). we can do it later, but for now at least put a TODO that we have to stop retrying after some number of retries.

iafuture (changes requested):

src/k2/persistence/plog_client/PlogClient.cpp
```diff
@@ -0,0 +1,203 @@
...
+                    break;
+            }
+            return seastar::make_ready_future<std::tuple<Status, uint32_t> >(std::tuple<Status, uint32_t>(std::move(return_status), std::move(sealed_offset)));
+        });
+}
+
+String PlogClient::_generatePlogId(){
+    String plogid = "TPCC_CLIENT_plog_0123456789";
+    std::mt19937 g(std::rand());
+    std::shuffle(plogid.begin()+18, plogid.end(), g);
```
iafuture, line 189: I've just realized this isn't a very good way to get a random id as it only produces permutations of [0-9]. I think we should pick a digit at random instead and set the value like you had before

src/k2/persistence/plog_client/PlogClient.h
```diff
@@ -0,0 +1,81 @@
...
+public:
```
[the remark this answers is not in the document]
piggesthjy: added

test/transport/PayloadTest.cpp
piggesthjy: added

src/k2/persistence/plog_service/PlogServer.h
piggesthjy: added

src/k2/persistence/plog_service/PlogServer.cpp
piggesthjy: changed the seal to be an immutable operation

src/k2/cpo/service/CPOService.cpp
```diff
@@ -302,6 +311,10 @@ String CPOService::_getCollectionPath(String name) {
     return _dataDir() + "/" + name + ".collection";
 }
 
+String CPOService::_getPersistenceClusterPath(String clusterName) {
+    return _dataDir() + "/" + clusterName;
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/cpo/service/CPOService.cpp
```diff
@@ -441,4 +454,40 @@ Status CPOService::_saveSchemas(const String& collectionName) {
...
+CPOService::handlePersistenceClusterGet(dto::PersistenceClusterGetRequest&& request) {
+    K2INFO("Received persistence cluster get request with name " << request.name);
+    auto cpath = _getPersistenceClusterPath(request.name);
+    dto::PersistenceCluster persistenceCluster;
+    Payload p;
+    if (!fileutil::readFile(p, cpath)) {
+        return RPCResponse(Statuses::S404_Not_Found("persistence cluster not found"), dto::PersistenceClusterGetResponse());
+    }
+    if (!p.read(persistenceCluster)) {
+        return RPCResponse(Statuses::S500_Internal_Server_Error("unable to read persistence cluster map data"), dto::PersistenceClusterGetResponse());
```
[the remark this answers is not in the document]
piggesthjy: fixed

src/k2/dto/PersistenceCluster.h
```diff
@@ -0,0 +1,67 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+    DAMAGES OR OTHER
+    LIABILITY,
+    WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+*/
+
+#include "Collection.h"
```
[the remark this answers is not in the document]
piggesthjy: removed

src/k2/dto/Persistence.h
```diff
@@ -0,0 +1,77 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+    DAMAGES OR OTHER
+    LIABILITY,
+    WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+*/
+
+#include "Collection.h"
```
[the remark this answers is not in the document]
piggesthjy: removed

src/k2/persistence/plog_client/PlogClient.cpp
piggesthjy: added the retry

test/plog/PlogTest.cpp
piggesthjy: added

src/k2/persistence/plog_client/PlogClient.cpp
iafuture: we decided to fix this in a later pr

src/k2/persistence/plog_client/PlogClient.cpp
iafuture: we decided to fix this in a later pr

iafuture (changes requested):

src/k2/persistence/plog_client/PlogClient.h
```diff
@@ -0,0 +1,89 @@
...
+    ~PlogClient();
+
+    // obtain the persistence cluster for a given name from the cpo and obtain the endpoints of the plog server
+    seastar::future<> init(String clusterName);
+
+    // allow users to select a specific persistence group by its name
+    bool selectPersistenceGroup(String name);
+
+    // create a plog with retry times
+    seastar::future<std::tuple<Status, String>> create(uint8_t retries = 1);
```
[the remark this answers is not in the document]
iafuture: we can revisit later if needed

iafuture (approved):

PR #32, opened 2020-08-27
issue22: 3SI Transaction Test cases

YazhiFeng:
Here are some questions encountered when implementing scenario-00:

Question 1:  When implementing a WRITE request, using the RPC().callRPC<>(Verbs::K23SI_WRITE, ...) operation causes "assertion error"  in the core code which throws an exception (the output is shown below). 
Therefore, the test code implements the WRITE request by first determining whether the collection exists. But I don't quite understand what WRITE is supposed to behave when the destination collection is not exist. 

<details>
<summary> assertion Error shown here</summary>

[0008:23:58:32.056.411]-3si_txn_test-(0) [ERROR] [/home/fengyazhi/chogori-platform/src/k2/dto/**Collection.cpp:127** @getPartitionForKey]Partition map does not begin with an empty string start key!
3si_txn_test: /home/fengyazhi/chogori-platform/src/k2/dto/Collection.cpp:127: k2::dto::PartitionGetter::PartitionWithEndpoint& k2::dto::PartitionGetter::getPartitionForKey(const k2::dto::Key&): Assertion `(it != _rangePartitionMap.begin())' failed.
Aborting on shard 0.
Backtrace:
  0x000000000016b8dc
  0x0000000000121328
  0x0000000000121622
  0x00000000001216e2
  0x000029ba744973bf
  /lib/x86_64-linux-gnu/libc.so.6+0x000000000004618a
  /lib/x86_64-linux-gnu/libc.so.6+0x0000000000025858
  /lib/x86_64-linux-gnu/libc.so.6+0x0000000000025728
  /lib/x86_64-linux-gnu/libc.so.6+0x0000000000036f35
  0x00000000002f1a43
  0x00000000000b5abb
  0x00000000000b9757
  0x000000000011d081
  0x000000000011d3c5
  0x0000000000152605
  0x00000000000fb741
  0x00000000000ca699
  0x0000000000076854
  /lib/x86_64-linux-gnu/libc.so.6+0x00000000000270b2
  0x000000000009140d
./test_3si_txn.sh: line 54: 40250 Aborted                 (core dumped) ./build/test/k23si/3si_txn_test --cpo_endpoint ${CPO} --k2_endpoints ${EPS} --enable_tx_checksum true --reactor-backend epoll --prometheus_port 63100  
</details>

Question 2: Because of Question 1, currently only WRITE verb is implemented in the test case. Should I implement all the verbs listed in the table e.g. READ, WRITE, PUSH, END, FINALIZE, HEARTBEAT ?

Question 3: Should I modify the core code according to the "Possible fix needed"? I don't know what to do for the “Possible fix checklist”.


jfunston:
1. For scenario00 tests, you cannot use the partition getter getPartitionForKey method to get the endpoint, because there is no partition map. You need to directly use the endpoint of the server node. Then the expected result is a timeout and no crash at the client or server.
2. Yes, we want it implemented for all of those verbs. The expected result is the same, a timeout.
3. No, for scenario00 you do not need to implement the possible fix.

YazhiFeng:
The test scenario-00 has been finished.

PR #34, opened 2020-09-02
issue22: 3SI Transaction Test Scenario-01

YazhiFeng:
issue22: 3SI Transaction Test Scenario-01


src/k2/module/k23si/Module.h
```diff
@@ -129,6 +129,29 @@ class K23SIPartitionModule {
         return result;
     }
 
+    // validate 2 Txns in PUSH requests are within the retention window for the collection. 
+    // return 0 if request is valid. 1 if incumbent mtr is stale, 2 if incubent mtr is stale, 3 if both are stale.
+    template <typename RequestT>
+    uint8_t _validatePushRetention(const RequestT& req) const {
+        bool result1 = req.incumbentMTR.timestamp.compareCertain(_retentionTimestamp) >= 0;
```
jfunston, line 136: We only need to validate that the challengerMTR is within the retention window. If the incumbentMTR is outside the retention window then the desired behavior is that the PUSH will go through and the incumbent will be evicted.
So we can also simplify the return of this _validate function to be a bool like the others.

src/k2/module/k23si/Module.h
```diff
@@ -129,6 +129,29 @@ class K23SIPartitionModule {
         return result;
     }
 
+    // validate 2 Txns in PUSH requests are within the retention window for the collection. 
+    // return 0 if request is valid. 1 if incumbent mtr is stale, 2 if incubent mtr is stale, 3 if both are stale.
+    template <typename RequestT>
+    uint8_t _validatePushRetention(const RequestT& req) const {
```
jfunston, line 135: This doesn't need to be a template since it can only be used with PUSH requests

src/k2/module/k23si/Module.h
```diff
@@ -129,6 +129,29 @@ class K23SIPartitionModule {
...
+            return 1;
+        } else {
+            return 0;
+        }
+    }
+
+    // validate keys in the requests must include non-empty partitionKey. return 0 if request parameter is valid
+    template <typename RequestT>
+    bool _validateRequestParameter(const RequestT& req) const {
+        return req.key.partitionKey.empty();
```
jfunston, line 152: Can we change this to return true if the request is valid, so that it matches the other _validate functions?

YazhiFeng:
Scenario-01 is modified according to the comments given from Ivan and the comments above. Please check it.

jfunston (approved):

PR #43, opened 2020-09-17
Implement scenarios 04-08. modified K23SI_testing.md

YazhiFeng:
Test cases of scenario 04&05 are designed and implemented which are listed below. 
The  K23SI_testing.md file is also modified.

### Test setup

- start a cluster and assign collection.
- write the following data, and keep the records in the given state

```
("SC04_pkey1","rKey1", v0) -> committed
```

### Test cases

- The test logic is shown below: 

image

| test case                                                    | Expected result                                   | Possible fix needed |
| ------------------------------------------------------------ | ------------------------------------------------- | ------------------- |
| Txn READ of data records before it begins                    | read prior value                                  |                     |
| Txn READ its own pending writes                              | read pending value                                |                     |
| Txn with earlier timestamp READ other txn's pending writes   | could not read pending value (read prior value)   |                     |
| Txn with older timestamp Read other txn's pending writes     | could not read pending value (NotFound)           |                     |
| Txn with earlier timestamp Read other txn's committed writes | could not read committed value (read prior value) |                     |
| Txn with older timestamp Read other txn's committed writes   | read committed value                              |                     |

### Test setup

- start a cluster and assign collection.
- write the following data, and keep the records in the given state

```
("SC05_pkey1","rKey1", v0) -> committed
```

### Test cases

| category                 | test case                                                    | Expected result                        | Possible fix needed |
| ------------------------ | ------------------------------------------------------------ | -------------------------------------- | ------------------- |
| fake-read-write conflict | Earlier Read txn encounters a WI, two txns have the same priority and tsoId | OK (read prior value)                  |                     |
| priority of Push()       | Txn with higher priority encounters a WI                     | created (challenger won the Push)      |                     |
| priority of Push()       | Txn with lower priority encounters a WI                      | AbortConflict (incumbent won the Push) |                     |
| timestamp of Push()      | Earlier WRITE Txn encounters a WI with the same priority     | AbortConflict (incumbent won the Push) |                     |
| timestamp of Push()      | Older READ/WRITE Txn encounters a WI with the same priority  | created (challenger won the Push)      |                     |
| tsoId of Push()          | WRITE Txn encounters a WI with the same priority and timestamp, but its tso ID is smaller | AbortConflict (incumbent won the Push) |                     |
| tsoId of Push()          | WRITE Txn encounters a WI with the same priority and timestamp, but its tso ID is bigger | created (challenger won the Push)      |                     |


YazhiFeng:
1. Test cases of scenario 06 are designed and implemented which are listed below. The returned status of 6th and 10th test cases is  S500_server caught exception processing request,  here may need  a fix. 

2. A bug in function doFinalize() is fixed.

## Scenario 06 - finalization

### Test setup

- start a cluster and assign collection. 
- Start a transaction that have not been End().

### Test cases


| test case                                                    | Expected result                                              | Possible fix needed |
| ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------- |
| Finalize is a non-exist record in this transaction           | OperationNotAllowed                                          |                     |
| Finalize_Commit partial record within this transaction       | OK                                                           |                     |
| Finalize a record whose status is commit                     | OperationNotAllowed                                          |                     |
| After Finalize_Commit partial record, other transactions read the record while this transaction is still in progress | OK                                                           |                     |
| After partial Finalization_commit, txn continues and then End_Commit all records | OK                                                           |                     |
| After partial Finalization_commit, txn continues and then End_Abort all records | S500_server caught exception processing request (partial finalize success) | need fix ？         |
| Finalize_Abort partial record within this transaction        | OK                                                           |                     |
| Record is read after it is Finalize_Abort  within the txn    | NotFound                                                     |                     |
| Finalize_abort a record who has already been finalized_abort | OK                                                           |                     |
| After partial Finalization_abort, txn continues and then End_Commit all records including the aborted records | S500_server caught exception processing request (partial finalize success) | need fix？          |
| After partial Finalization_abort, txn continues and then End_Abort all records | OK                                                           |                     |
| The TRH and MTR parameters of Finalize do not match          | OperationNotAllowed                                          |                     |
| During async end_abort interval, finalize_commit partial keys, validate the keys with READ | OK (finalize_commit keys can be read)                        |                     |

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1797,690 @@ seastar::future<> testScenario02() {
...
+        );
+    })
+    .then([]{
+        K2INFO("Scenario 04 setup done.");
+        return getTimeNow();
+    })
+    .then([this](dto::Timestamp&& ts) {
+        std::cout << std::endl;
+
+        return seastar::do_with(
```
jfunston, line 1842: These more complicated scenarios are getting a bit hard to follow in the code, especially with the nesting. Maybe you can refactor the scenario into a class and use member variables instead of do_with clauses?

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1797,690 @@ seastar::future<> testScenario02() {
...
+                            K2INFO("Value in Txn(A): " << val1);
+                            K2INFO("Value in Txn(B): " << val2);
+                            K2INFO("Txn(B)::OP_write_intent_k2_v1. " << "status: " << status3.code << " with MESG: " << status3.message);
+                            K2EXPECT(status1, dto::K23SIStatus::OK);
+                            K2EXPECT(status2, dto::K23SIStatus::OK);
+                            K2EXPECT(status3, dto::K23SIStatus::Created);
+                            K2EXPECT(val1, v1);
+                            K2EXPECT(val2, v0);
+                        })
+                        .then([&] {
```
jfunston, line 1902: Here we can use the K23SIInspect RPCs to make the test a bit more robust. For example you can use K23SIInspectRecordsRequest to check that the write intent still exists on the server after it is read. https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/dto/K23SIInspect.h

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1797,690 @@ seastar::future<> testScenario02() {
...
+                })
+                .then([&] {
+                    K2INFO("------- SC05.case2 (Txn with higher priority encounters a WI) -------");
+                    return doWrite(k1, v1, prioHighMtr, k1, collname, false, true, ErrorCaseOpt::NoInjection)
+                    .then([](auto&& response) {
+                        auto& [status, resp] = response;
+                        K2INFO("SC05.case02::OP_push_prioHighMtr. " << "status: " << status.code << " with MESG: " << status.message);
+                        K2EXPECT(status, dto::K23SIStatus::Created);
+                    });
+                })
```
jfunston, line 2055: Same for these tests, we can use the K23SIInsepect RPCs to check that that the aborted transaction is in the force aborted state.

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1797,690 @@ seastar::future<> testScenario02() {
...
+            .then([&] {
+                return doFinalize(k4, k4, mtr, collname, false, ErrorCaseOpt::NoInjection)
+                .then([](auto&& response)  {
+                    auto& [status, val] = response;
+                    K2INFO("SC06.case11::OP_finalize_abort_k4. " << "status: " << status.code << " with MESG: " << status.message);
+                    K2EXPECT(status, dto::K23SIStatus::OK);
+                });
+            })
+            .then([&] {
+                return doWrite(k6, v0, mtr, k4, collname, false, false, ErrorCaseOpt::NoInjection)
```
jfunston, line 2395: Let's remove this write, then the test is a more realistic scenario where the abort-finalize is sent more than once as part of retry logic.

docs/RFC/K23SI_testing.md
```diff
@@ -76,10 +76,88 @@ Notes:
...
+| ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------- |
+| Finalize is a non-exist record in this transaction           | OperationNotAllowed                                          |                     |
+| Finalize_Commit partial record within this transaction       | OK                                                           |                     |
+| Finalize a record whose status is commit                     | OperationNotAllowed                                          |                     |
+| After Finalize_Commit partial record, other transactions read the record while this transaction is still in progress | OK                                                           |                     |
+| After partial Finalization_commit, txn continues and then End_Commit all records | OK                                                           |                     |
+| After partial Finalization_commit, txn continues and then End_Abort all records | S500_server caught exception processing request (partial finalize success) | need fix ？         |
```
jfunston, line 153: You can add a note here that it is expected that the data will be in an inconsistent state. An S500 is ok as long as the server is not crashing and can continue to process new transactions.

docs/RFC/K23SI_testing.md
```diff
@@ -76,10 +76,88 @@ Notes:
...
+| Finalize_Abort partial record within this transaction        | OK                                                           |                     |
+| Record is read after it is Finalize_Abort  within the txn    | NotFound                                                     |                     |
+| Finalize_abort a record who has already been finalized_abort | OK                                                           |                     |
+| After partial Finalization_abort, txn continues and then End_Commit all records including the aborted records | S500_server caught exception processing request (partial finalize success) | need fix？          |
```
jfunston, line 157: Same as above comment.

jfunston (changes requested):

test/k23si/3SITxnTest.cpp
YazhiFeng: A test logic figure for  scenario 04 is added. And I refactor the implementation to make it more easier to read. To keep the style consistent with the rest of the scenarios, I did not refactor it into a class and member variables.

test/k23si/3SITxnTest.cpp
YazhiFeng: K23SIInspect RPCs are added and used in test cases.

YazhiFeng:
scenario 04-06 is modified according to comments.

1. Scenario 04 is refactored to make it clear to read. A test logic figure is also added.

2. K23SIInspect RPCs are used in the test cases.

3. Other comments are all modified accordingly.

YazhiFeng:
Scenario 07 implementation is also included in the PR.

YazhiFeng:
Scenario 08 implementation is added in the PR.

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1822,1277 @@ seastar::future<> testScenario02() {
...
+                .then([&](auto&& response)  {
+                    auto& [status, val] = response;
+                    K2INFO("SC08.case01::OP_inspect_txn_1. " << "status: " << status.code << " with MESG: " << status.message);
+                    K2INFO("State of txn_1: " << val.state);
+                    K2EXPECT(status, dto::K23SIStatus::OK);
+                    K2EXPECT(val.state, dto::TxnRecordState::ForceAborted);
+                });
+            })
+            .then([&] {
+                K2INFO("------- SC08.case02 ( Txn WRITE happens before the Read-Cache record read time ) -------");
```
jfunston, line 2973: You should do an END (with commit = false and syncFinalize = true) first to clean up the first write before doing this test case. Then you would expect AbortConflict on the response (as it is now) and InProgress on the txn inspect

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1822,1277 @@ seastar::future<> testScenario02() {
...
+                .then([&] {
+                    return doInspectTxn(k1, mtr1, collname)
+                    .then([&](auto&& response)  {
+                        auto& [status, val] = response;
+                        K2INFO("SC08.case02::OP_inspect_txn_1. " << "status: " << status.code << " with MESG: " << status.message);
+                        K2INFO("State of txn_1: " << val.state);
+                        K2EXPECT(status, dto::K23SIStatus::OK);
+                        K2EXPECT(val.state, dto::TxnRecordState::ForceAborted);
+                    });
+                });
```
jfunston, line 2989: Need another END here

test/k23si/3SITxnTest.cpp
```diff
@@ -1794,10 +1822,1277 @@ seastar::future<> testScenario02() {
...
+            });
+        }); // end do-with
+    }); // end sc-07
+}
+
+seastar::future<> testScenario08() {
+    std::cout << std::endl << std::endl;
+    K2INFO("+++++++ TestScenario 08: server-initiated txn abort +++++++");
```
jfunston, line 2930: A transaction should only be in the forceAborted state if it was the incumbent in a PUSH operation and lost (or it failed heartbeat or retention window, but those are not relevant here). If a challenger transaction loses a PUSH, the client will see an AbortConflict but the transaction will be in the InProgress state and we rely on the client will END with an abort. If we don't see that then it is a bug in the code or the test. Here you are missing some END requests so the state of the previous test cases are not cleaned up. I commented on a couple but you can see if there are more.

jfunston (changes requested):

YazhiFeng:
Scenario 08 is modified according to the comments. 
END is performed to make the transaction in the TxnRecordState::Created state. 
Expected result in K23SI_testing.md is also modified accordingly.

jfunston (approved):

PR #53, opened 2020-09-23
SKV Query interface and dto

jfunston:


docs/SKV.md
```diff
@@ -114,3 +114,32 @@ with read or write requests because the SKV Client does not know if the user wil
...
+
+
+Query predicates consist of a field name, an expected field type, a relational operator, and a literal 
+of the field type. For example, {"AGE", Int, LESS\_THAN, 25} represents a predicate that 
+selects for records where the AGE field is less than 25. By expressing predicate fields by name rather 
+than positionally we can support queries across schema versions more easily. Similarly, projection is 
+also expressed by field name. Predicates are logically ANDed together by the server.
+
+
+The result of a Query is a set of SKVRecords, which is paginated. The server decides how many records to
```
iafuture, line 134: Could the result be an aggregate (sum, count, avg, etc?)

docs/SKV.md
```diff
@@ -114,3 +114,32 @@ with read or write requests because the SKV Client does not know if the user wil
...
+An SKV Query is a read-only scan operation with predicates and projection. It operates over a single 
+schema (but over all versions of that schema, see schema\_versioning.md for more details). The SKV 
+Query interface is a fixed-function API, not generic pushdown.
+
+
+Query predicates consist of a field name, an expected field type, a relational operator, and a literal 
+of the field type. For example, {"AGE", Int, LESS\_THAN, 25} represents a predicate that 
+selects for records where the AGE field is less than 25. By expressing predicate fields by name rather 
+than positionally we can support queries across schema versions more easily. Similarly, projection is 
+also expressed by field name. Predicates are logically ANDed together by the server.
```
iafuture, line 131: any way we can leave it up to caller to specify the boolean aggregation (AND, OR, XOR, NOT etc). We should think if we can allow nested composition e.g. ( (age BETWEEN 20,30 AND  order_count > 30) OR (age > 30) )

src/k2/dto/K23SI.h
```diff
@@ -219,6 +219,44 @@ struct K23SIWriteResponse {
...
+    Payload predicateOperands;
+
+    std::vector<String> projection; // Fields by name to include in projection
+
+    K2_PAYLOAD_FIELDS(pvid, collectionName, mtr, key, recordLimit, includeVersionMismatch, predicateOps,
+            predicateFields, predicateTypes, predicateOperands, projection);
+};
+
+struct K23SIQueryResponse {
+    std::vector<SKVRecord::Storage> records;
```
iafuture, line 256: continuation token?

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -98,6 +95,39 @@ class K23SIClientConfig {
 
 class K2TxnHandle;
 
+// Represents a new or in-progress query (aka read scan with predicate and projection)
+class Query {
+public:
+    Query() = default;
+
+    template <typename T>
+    void addPredicate(dto::K23SIQueryOp op, T operand, const String& fieldName);
```
iafuture, line 104: we should probably support N-ary predicates via T&&... operands

jfunston:
I haven't traced the full execution yet, but what I can tell from yugabytedb is that they support a range scan on keys with a single conditional clause on the documentDB:
https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/docdb/doc_pgsql_scanspec.h#L26
https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/common/pgsql_protocol.proto#L43
https://github.com/yugabyte/yugabyte-db/blob/master/src/yb/common/common.proto#L629

Issue #44, opened 2020-09-21
Implement SKV Partial Update


jfunston:
A partial update is when the user only wants to write a subset of the fields of a record. We want to optimize for this case with a new partialUpdate RPC and interface, which will reduce traffic over the network.

- Implement a new DTO and RPC for partialUpdate based on the SKV write RPC.
- Implement the RPC in the 3SI Module. For this task we do not need to store delta records on the server, the server can copy and create full records. We can have a separate task for storing delta records.
- Implement the interface in https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/module/k23si/client/k23si_client.h The interface should have two overloaded functions, one for specifying fields by  name and one by index: "seastar::future\<WriteResult\> partialUpdate(T& record, std::vector<k2::String> fieldsToUpdate" and "seastar::future\<WriteResult\> partialUpdate(T& record, std::vector<uint32_t> fieldsToUpdate"
- A more complicated part of this task is dealing with schema versioning. See the design here: https://github.com/futurewei-cloud/chogori-platform/blob/master/docs/RFC/schema_versioning.md#partial-updates
- Add tests, including tests for the schema versioning cases mentioned above, here: https://github.com/futurewei-cloud/chogori-platform/blob/master/test/k23si/SKVClientTest.cpp
- Modify the TPC-C test to use partial updates. If you have trouble running TPC-C for testing let me know and I can help or take over this part of the task. https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/cmd/tpcc/transactions.h

YazhiFeng:
Hi, Justin. I got some questions about implementing partialUpdate. 

1. When I implement the interface in k23si_client.h, considering a user app sending partialUpdate requests through the k23_client, the most common way to call this partialUpdate interface I think is : The user declares a SKVRecord and enters updated fields using serializeNext, skips non-updated fields using skipNext, and then calls the interface with txnHandle.partialUpdate <>(record) from k23si_client. Should I implement interface in this scenario?

So I don't know under what circumstances this kind of interface "partialUpdate(T& record, std::vector<> fieldsToUpdate)" is called. Should every fields in the record be serialized? 

2. I want to confirm that does partialUpdate use SKVRecord::Storage to send value to server as K23SIWriteRequest does.  But it can reduce network traffic through less fieldData. 

3. Is the partialUpdate relative to the latest record? It means that I can supplement the non-updated fields with the fields from the latest committed record.

[...]

PR #54, opened 2020-09-30
K23SI Partial Update

YazhiFeng:
Hi, Justin:
1. Partial update functionality and test cases in SKVClientText have been implemented. 

2. But still leave some problems: 
    1). The WI bug we discuss in https://github.com/futurewei-cloud/chogori-platform/issues/44 was not well fixed. Considering the logical fitness of the entire k23SI module, I may leave that to you.
    2). The TPC-C test has not been modified yet. I will look into it but maybe I do not enough RDMA machines to do the performance test. I'll see what I can do.

### DIFF
--- a/src/k2/dto/K23SI.h
+++ b/src/k2/dto/K23SI.h
@@ -219,6 +219,32 @@ struct K23SIWriteResponse {
     K2_PAYLOAD_EMPTY;
 };
 
+struct K23SIPartialUpdateRequest{
+    Partition::PVID pvid; // the partition version ID. Should be coming from an up-to-date partition map
+    String collectionName; // the name of the collection
+    K23SI_MTR mtr; // the MTR for the issuing transaction
+    // The TRH key is used to find the K2 node which owns a transaction. It should be set to the key of
+    // the first write (the write for which designateTRH was set to true)
+    // Note that this is not an unique identifier for a transaction record - transaction records are
+    // uniquely identified by the tuple (mtr, trh)
+    Key trh;
+    bool isDelete = false; // is this a delete write?
+    bool designateTRH = false; // if this is set, the server which receives the request will be designated the TRH
+    // use the name "key" so that we can use common routing from CPO client
+    Key key; // the key for the write
+    SKVRecord::Storage value; // the value of the write
+    K2_PAYLOAD_FIELDS(pvid, collectionName, mtr, trh, isDelete, designateTRH, key, value);
+    friend std::ostream& operator<<(std::ostream& os, const K23SIPartialUpdateRequest& r) {
+        return os << "{pvid=" << r.pvid << ", colName=" << r.collectionName
+                  << ", mtr=" << r.mtr << ", trh=" << r.trh << ", key=" << r.key << ", isDelete="
+                  << r.isDelete << ", designate=" << r.designateTRH << "}";
+    }
+};
+
+struct K23SIPartialUpdateResponse {
+    K2_PAYLOAD_EMPTY;
+};
+
 struct K23SITxnHeartbeatRequest {
     // the partition version ID for the TRH. Should be coming from an up-to-date partition map
     Partition::PVID pvid;
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -64,6 +64,10 @@ seastar::future<> K23SIPartitionModule::start() {
         return handleWrite(std::move(request), dto::K23SI_MTR_ZERO, FastDeadline(_config.writeTimeout()));
     });
 
+    RPC().registerRPCObserver<dto::K23SIPartialUpdateRequest, dto::K23SIPartialUpdateResponse>(dto::Verbs::K23SI_PARTIAL_UPDATE, [this](dto::K23SIPartialUpdateRequest&& request) {
+        return handlePartialUpdate(std::move(request), dto::K23SI_MTR_ZERO, FastDeadline(_config.writeTimeout()));
+    });
+
     RPC().registerRPCObserver<dto::K23SITxnPushRequest, dto::K23SITxnPushResponse>
     (dto::Verbs::K23SI_TXN_PUSH, [this](dto::K23SITxnPushRequest&& request) {
         return handleTxnPush(std::move(request));
@@ -232,7 +236,8 @@ K23SIPartitionModule::handleRead(dto::K23SIReadRequest&& request, dto::K23SI_MTR
     return _makeReadOK(versions.begin() == versions.end() ? nullptr : &(versions[0]));
 }
 
-bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request, std::deque<dto::DataRecord>& versions) {
+template <typename RequestT>
+bool K23SIPartitionModule::_validateStaleWrite(const RequestT& request, std::deque<dto::DataRecord>& versions) {
     if (!_validateRetentionWindow(request)) {
         // the request is outside the retention window
         return false;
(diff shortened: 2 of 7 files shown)


YazhiFeng:
The test scenarios is listed below:

## Scenario 04 - Partial update with the same schema version

### Test setup

- start a cluster and write the basic version of data.
- _client begins a TXN, and gets the schema version.

### Test cases

| test case                                                    | Expected result | Note                                                         |
| ------------------------------------------------------------ | --------------- | ------------------------------------------------------------ |
| Partial update value fields, with Skipped PartitionKey and RangeKey fields | BadParameter    | missing partition key                                        |
| Partial update value fields, including PartitionKey and RangeKey fields | KeyNotFound     | Equals to write another key. Cannot use it to create a new SKVRecord. |
| Partial update every value， using field index to indicate the fieldsToUpdate | OK              | k23si_client do not encode payload in this case. It is a quick path. |
| Partial update some value fields using field name to indicate the fieldsToUpdate | OK              |                                                              |
| Partial update some fields with NULL fieldsToUpdate          | OK              | NULL fieldsToUpdate implies copy of the previous data. All fields are not updated |
| Partial update while the sematic of excludedFields in record is different from fieldsToUpdate. <br /><br />1): fieldsToUpdate indicate a field shall be updated, but this field is skipped in the record | BadParameter    | must got a value while updating a field                      |
| 2): fieldsToUpdate indicate  a field shall not be updated, but record has a value of this field | OK              | the value in record would be dropped                         |



## Scenario 05 - Partial update with different schema versions

### Test setup

- start a cluster and write the basic version of data.
- insert more schema versions.
- _client begins a TXN, and gets the schema versions.

### Test cases

| test case                                                    | Expected result | Note |
| ------------------------------------------------------------ | --------------- | ---- |
| same fields but different orders, copy pervious data (null fieldsToUpdate) | OK              |      |
| missing a field with same field name but different field type | badParameter    |      |
| add some fields, but leaving them not updated                | badParameter    |      |
| add some fields, do not contain some of the pre-existing fields | OK              |      |
| decrease some fields, and update some fields                 | OK              |      |

src/k2/module/k23si/Module.cpp
```diff
@@ -272,6 +277,227 @@ bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request,
...
+    return true;
+}
+
+bool K23SIPartitionModule::_parsePartialRecord(dto::SKVRecord::Storage& storage, const k2::String& schemaName, std::deque<dto::DataRecord>& versions) {
+    auto schemaIt = _schemas.find(schemaName);
+    if (schemaIt == _schemas.end()) return false;
+    auto schemaVer = schemaIt->second.find(storage.schemaVersion);
+    if (schemaVer == schemaIt->second.end()) return false;
+    dto::Schema schema = schemaVer->second;
+    Payload payload([&] { return Binary(4096); });
```
jfunston, line 400: You can use Payload::DefaultAllocator here

src/k2/module/k23si/Module.cpp
```diff
@@ -272,6 +277,227 @@ bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request,
...
+            } // else, base SKVRecord skipped this field, and this request does not update this field, do nothing
+        }
+        storage.fieldData = std::move(payload);
+        storage.fieldData.truncateToCurrent();
+    } else { 
+        // slow path--different schema version. 
+        auto latestSchemaVer = schemaIt->second.find(versions[0].value.schemaVersion);
+        if (latestSchemaVer == schemaIt->second.end()) return false;
+        dto::Schema baseSchema = latestSchemaVer->second;
+        std::unordered_map<std::size_t, k2::String> mapString; // unordered_map temporarily stores base SKVRecords
```
jfunston, line 424: Is it possible to do this without going to and from a map?

src/k2/module/k23si/Module.cpp
```diff
@@ -358,6 +593,111 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, dto::K23SI_M
     });
 }
 
+seastar::future<std::tuple<Status, dto::K23SIPartialUpdateResponse>>
+K23SIPartitionModule:: handlePartialUpdate(dto::K23SIPartialUpdateRequest&& request, dto::K23SI_MTR sitMTR, FastDeadline deadline) {
```
jfunston, line 597: Is it possible to reduce some of the code duplication here with the handleWrite function?

jfunston (changes requested):

src/k2/dto/K23SI.h
```diff
@@ -219,6 +219,32 @@ struct K23SIWriteResponse {
...
+struct K23SIPartialUpdateRequest{
+    Partition::PVID pvid; // the partition version ID. Should be coming from an up-to-date partition map
+    String collectionName; // the name of the collection
+    K23SI_MTR mtr; // the MTR for the issuing transaction
+    // The TRH key is used to find the K2 node which owns a transaction. It should be set to the key of
+    // the first write (the write for which designateTRH was set to true)
+    // Note that this is not an unique identifier for a transaction record - transaction records are
+    // uniquely identified by the tuple (mtr, trh)
+    Key trh;
+    bool isDelete = false; // is this a delete write?
```
jerryhfeng, line 231: Delete should be just a WriteRequest and partialUpdate doesn't need to support it, right?

src/k2/module/k23si/Module.cpp
```diff
@@ -272,6 +277,227 @@ bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request,
     return true;
 }
 
+std::size_t K23SIPartitionModule::_findField(const dto::Schema schema, k2::String fieldName ,dto::FieldType fieldtype) {
```
jerryhfeng, line 280: Not related with new code, but 10 lines above, we have
"else if (versions.size() > 1 && versions[0].status == dto::DataRecord::WriteIntent &&
        request.mtr.timestamp.compareCertain(versions[1].txnId.mtr.timestamp) <= 0) {"
This optimization is good/correct, but if re-write of same KV in the same transaction is not over-write the same dataRecord in the Deque, this line will cause the re-write fail on the third time in a transaction. A bug scenario Yazhi mentioned in previous meeting.

src/k2/module/k23si/Module.cpp
```diff
@@ -350,6 +578,13 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, dto::K23SI_M
                 });
         }
     }
+    
+    // Clean up if req and WI are the same transaction
```
jerryhfeng, line 582: This should fix the bug of write same record in a transaction three or more times. But let's move it into above    
 if (versions.size() > 0 && versions[0].status == dto::DataRecord::WriteIntent) { 
Remove previous write in memory if it is written same transaction, otherwise to the rest of check including
if (sitMTR == rec.txnId.mtr) {
            K2DEBUG("Partition: " << _partition << ", post-push winner for key " << request.key);
            // this is a post-PUSH request which won over the siting WI and we still have the WI in cache
            _queueWICleanup(std::move(rec));
            versions.pop_front();
        }
        else if (rec.txnId.mtr != rqmtr) {

Also, I suggest that we can optimize this function related with push. First remove the parameter sitMtr from the API, it is kind of confusing. Then inside request (or mtr cache when we have one), we can keep a list of txnId(key) this request transaction already wins through push, so that we can check against this list first when we need to push, instead of using sitMtr for it. This way, assuming the list was maintained on the request transaction, we can avoid push against other writeintent of the transactions we already pushed and won before. This optimization can be done later and I will sync with Ivan about this as well.

src/k2/dto/K23SI.h
jfunston: yes, I agree

jfunston:
If you want to save the TPC-C changes for the next PR that is fine. You can use test/integration/test_k23si_tpcc.sh on your local VM to test if the changes work. For performance testing, I recently updated the scripts and README in the cluster directory. You will need to modify the configuration to work on 3 machines, with fewer cores used and fewer TPC-C warehouses used. You don't need to do it for this task but getting setup for performance testing will be very useful soon.

src/k2/dto/K23SI.h
```diff
@@ -219,6 +219,32 @@ struct K23SIWriteResponse {
...
+    bool designateTRH = false; // if this is set, the server which receives the request will be designated the TRH
+    // use the name "key" so that we can use common routing from CPO client
+    Key key; // the key for the write
+    SKVRecord::Storage value; // the value of the write
```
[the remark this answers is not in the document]
YazhiFeng: k23si client has been modified according to comments:

1. fieldsToUpdate is added in the K23SIPartialUpdateRequest. And partialUpdate() in k23si_client has been re-write.
2. partialUpdate function is overloaded. 
3. Now client only make and send partialUpdate request. Other operations have been removed.
4. 'delete' parameter in K23SIPartialUpdateRequest is removed.

src/k2/module/k23si/Module.cpp
YazhiFeng: k23si Module has been modified: 

1. handlePartialUpdate was not combined with handleWrite function. Because I think two different functions can make the code clearer to read.
2. _parsePartialRecord function was re-written. _makeFieldsForSameVersion function is added to make full-record-WI in the case of same schema and same version. _makeFieldsForDiffVersion function is added to make full-record-WI in the case of same schema and different versions.
3. unordered_map has been removed. Instead, use an vector to record the offset of each field in base SKVRecord (in function _makeFieldsForDiffVersion ).

also, added two more test cases for SKVClientTest and they are all listed below:

### Test cases for same schema version update

| test case                                                    | Expected result | Note                                                         |
| ------------------------------------------------------------ | --------------- | ------------------------------------------------------------ |
| Partial update value fields, with Skipped PartitionKey and RangeKey fields | BadParameter    | missing partition key                                        |
| Partial update value fields, including PartitionKey and RangeKey fields | KeyNotFound     | Equals to write another key. Cannot use it to create a new SKVRecord. |
| Partial update every value， using field index to indicate the fieldsToUpdate | OK              | k23si_client do not encode payload in this case. It is a quick path. |
| Partial update some value fields using field name to indicate the fieldsToUpdate | OK              |                                                              |
| Partial update some fields with NULL fieldsToUpdate          | OK              | NULL fieldsToUpdate implies copy of the previous data. All fields are not updated |
| Partial update while the sematic of excludedFields in record is different from fieldsToUpdate. <br /><br />1): fieldsToUpdate indicate a field shall be updated, but this field is skipped in the record | BadParameter    | must got a value while updating a field                      |
| 2): fieldsToUpdate indicate  a field shall not be updated, but record has a value of this field | OK              | the value in record would be dropped                         |
| Updates an null field while keeping other null fields remain null | OK              |                                                              |

### Test cases for different schema versions update

| test case                                                    | Expected result | Note |
| ------------------------------------------------------------ | --------------- | ---- |
| same fields but different orders, copy pervious data (null fieldsToUpdate) | OK              |      |
| missing a field with same field name but different field type | badParameter    |      |
| add some fields, but leaving them not updated                | badParameter    |      |
| add some fields, do not contain some of the pre-existing fields | OK              |      |
| decrease some fields, and update value-fields to null-fields | OK              |      |
| update fields from null to null                              | OK              |      |

YazhiFeng:
PartialUpdate function modules have been modified according to the comments. But TPCC testing is not yet complete.

src/k2/module/k23si/Module.cpp
```diff
@@ -272,6 +277,285 @@ bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request,
...
+    auto schemaVer = schemaIt->second.find(request.value.schemaVersion);
+    if (schemaVer == schemaIt->second.end()) return false;
+    dto::Schema schema = schemaVer->second;
+    
+    if (!request.value.excludedFields.size()) {
+        request.value.excludedFields = std::vector<bool>(schema.fields.size(), false);
+    }
+    
+    Payload basePayload = versions[0].value.fieldData.shareAll();   // base payload
+    Payload payload(Payload::DefaultAllocator);                     // payload for new record
```
jfunston, line 540: I think payload and basePayload are unused here and can be removed?

src/k2/module/k23si/Module.cpp
```diff
@@ -526,6 +921,22 @@ K23SIPartitionModule::_createWI(dto::K23SIWriteRequest&& request, std::deque<dto
     return _persistence.makeCall(versions.front(), deadline);
 }
 
+seastar::future<>
+K23SIPartitionModule::_createWI(dto::K23SIPartialUpdateRequest&& request, std::deque<dto::DataRecord>& versions, FastDeadline deadline) {
+    K2DEBUG("Partition: " << _partition << ", PartialUpdate Request creating WI: " << request);
+    dto::DataRecord rec;
+    rec.key = std::move(request.key);
+    // we need to copy this data into a new memory block so that we don't hold onto and fragment the transport memory
+    rec.value = request.value.copy();
```
jfunston, line 930: You should explicitly set rec.isTombstone = false here

src/k2/module/k23si/Module.cpp
```diff
@@ -272,6 +277,285 @@ bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request,
...
+    request.value.fieldData.truncateToCurrent();
+    return true;
+}
+
+bool K23SIPartitionModule::_parsePartialRecord(dto::K23SIPartialUpdateRequest& request, std::deque<dto::DataRecord>& versions) {
+    auto schemaIt = _schemas.find(request.key.schemaName);
+    if (schemaIt == _schemas.end()) return false;
+    auto schemaVer = schemaIt->second.find(request.value.schemaVersion);
+    if (schemaVer == schemaIt->second.end()) return false;
+    dto::Schema schema = schemaVer->second;
```
jfunston, line 533: You can use dto::Schema& to avoid a memory copy here

src/k2/module/k23si/Module.cpp
```diff
@@ -350,14 +642,117 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, dto::K23SI_M
...
+                });
+        }
+    }
+
+    // parse the partial record to full record
+    if ( !versions.size() ) {
+        // cannot parse partial record without a version
+        return RPCResponse(dto::K23SIStatus::KeyNotFound("can not partial update without a version"), dto::K23SIPartialUpdateResponse{});
+    }
+    if (!_parsePartialRecord(request, versions)) {
```
jfunston, line 736: You should check if versions[0] is a tombstone or not before calling this function. If it is a tombstone then you can return a KeyNotFound status.

src/k2/module/k23si/Module.cpp
```diff
@@ -272,6 +277,285 @@ bool K23SIPartitionModule::_validateStaleWrite(dto::K23SIWriteRequest& request,
...
+            }
+        }        
+    }
+    
+    request.value.fieldData = std::move(payload);
+    request.value.fieldData.truncateToCurrent();
+    return true;
+}
+
+bool K23SIPartitionModule::_makeFieldsForDiffVersion(dto::Schema schema, dto::Schema baseSchema, dto::K23SIPartialUpdateRequest& request, std::deque<dto::DataRecord>& versions) {
```
jfunston, line 432: You can pass the schemas by reference to avoid a memory copy, and it might be cleaner to pass a reference to the first record version instead of passing the whole deque here. Same comment for _makeFieldsForSameVersion above

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -309,6 +336,70 @@ class K2TxnHandle {
...
+        }
+        _client->write_ops++;
+        _ongoing_ops++;
+
+        dto::K23SIPartialUpdateRequest* request = nullptr;
+        if constexpr (std::is_same<T1, dto::SKVRecord>()) {
+            request = makePartialUpdateRequest(record, fieldsToUpdate);
+        } else {
+            SKVRecord skv_record(record.collectionName, record.schema);
+            record.__updateFields(skv_record);
```
jfunston, line 375: This line should be: record.__writeFields(skv_record);

test/k23si/SKVClientTest.cpp
```diff
@@ -274,6 +276,585 @@ seastar::future<> runScenario03() {
...
+                                K2EXPECT(response.status, k2::dto::K23SIStatus::OK);
+                                
+                                std::optional<k2::String> partkey = response.value.deserializeNext<k2::String>();
+                                std::optional<k2::String> rangekey = response.value.deserializeNext<k2::String>();
+                                std::optional<k2::String> data1 = response.value.deserializeNext<k2::String>();
+                                std::optional<k2::String> data2 = response.value.deserializeNext<k2::String>();
+                                K2EXPECT(*partkey, "partkey_s04");
+                                K2EXPECT(*rangekey, "rangekey_s04");
+                                K2EXPECT(*data1, "data1_v3");   // "f1" field was updated to v3 
+                                K2EXPECT(*data2, "");           // "f2" field was updated to null
```
jfunston, line 509: It would be better to use K2EXPECT(data2, false) in cases like this where you expect a null field.

jfunston (changes requested):

test/k23si/SKVClientTest.cpp
YazhiFeng: K2EXPECT(data2, false) cannot compile，while K2ASSERT(!data2, "msg") can. But assert doesn't look so pretty here.

src/k2/module/k23si/Module.cpp
YazhiFeng: yes, they should be removed.

src/k2/module/k23si/Module.cpp
YazhiFeng: Check for tombstone is added and put together with the previous if() statement

```
if ( !versions.size() || versions[0].isTombstone) {
        // cannot parse partial record without a version
        return RPCResponse(dto::K23SIStatus::KeyNotFound("can not partial update with no/deleted version"), dto::K23SIPartialUpdateResponse{});
    }
```

src/k2/module/k23si/Module.cpp
YazhiFeng: yes,` rec.isTombstone = false` is added.

jfunston (approved):
Looks good, thanks! I made a new task for adding partial update to TPC-C, you can start on it next: https://github.com/futurewei-cloud/chogori-platform/issues/61

PR #65, opened 2020-10-15
Filter expressions

iafuture:


src/k2/dto/Expression.cpp
```diff
@@ -0,0 +1,298 @@
...
+        }
+    }
+    Value& val;
+    SKVRecord& rec;
+    FieldType type = FieldType::NOT_KNOWN;
+    bool nullLast = false;
+    int sfieldIndex = -1;
+
+    template <typename T>
+    std::optional<std::tuple<bool, T>> get() {
```
jfunston, line 26: Shouldn't this be std::tuple<bool, std::optional<T>> ?

src/k2/dto/Expression.cpp
```diff
@@ -0,0 +1,298 @@
...
+    if (std::get<1>(*a) == std::get<1>(*b)) {
+        return -1;
+    }
+    if (std::get<1>(*a) > std::get<1>(*b)) {
+        return 1;
+    }
+    return -1;
+}
+
+#define CAST_APPLY_VALUE(func, a, ...)              \
```
jfunston, line 104: This is fairly generic now which is great. As one of my next tasks I can find all the places in the codebase that has a switch like this and consolidate.

src/k2/dto/Expression.cpp
```diff
@@ -0,0 +1,298 @@
+#include "Expression.h"
```
jfunston, line 1: Is it possible to move this out of dto? The code in this .cpp is only needed by the server, but I guess it would be weird to separate the .h and .cpp.

src/k2/dto/SKVRecord.h
```diff
@@ -90,7 +90,7 @@ class SKVRecord {
             }
         }
 
-        throw new std::runtime_error("Schema not followed in record deserialization");
+        return std::nullopt;
```
jfunston, line 93: Do we still need this change? If we keep it, what about the case where the user tries to deserialize with the wrong type, should that also return a nullopt?

jerryhfeng: Seems the change is reflected in the test case code as well. But not sure why we need this behavior change as well. Could you explain either online or offline?

src/k2/dto/Expression.h
```diff
@@ -0,0 +1,106 @@
...
+    ENDS_WITH,    // value operands only
+    AND,          // mixed value and/or nested expression operands
+    OR,           // mixed value and/or nested expression operands
+    XOR,          // mixed value and/or nested expression operands
+    NOT,           // mixed value and/or nested expression operands
+    UNKNOWN
+};
+
+// A Value in the expression model. It can be either
+// - a field reference which sets the fieldName, or
```
jerryhfeng, line 40: Should we consider using schema Id + field Idx instead of fieldName? Since when the client construct the col/field reference, it has already resolved the column/field name with schema, so using fieldName still will be less efficient as it need to be resolved again, as in SchematizedValue constructor?

src/k2/dto/SKVRecord.h
iafuture: This is needed to support our advertised "all fields are optional" interface. Imagine these scenarios in which we need to decide if we should throw or not.
0. Key is in schema, field is set but it is of incorrect type
1. Key is in schema but field is not set
2. Key is not in schema

I propose for 0 we throw since we have this key, but the type doesn't match.
For 1. and 2. it makes more sense to me to return std::nullopt to signal "this key is not set". whether it exists in the schema or not, it doesn't change the fact that it "isn't set"

Coincidentally, :D this also makes my code a bit easier since I don't have to handle case 2 specially. With a scan filter, I have to be able to evaluate a key condition (e.g. "age" > 20) on records which
1. have the key and is of compatible type
2. have the key but it isn't of compatible type
3. have the key in schema but it isn't set
4. don't have the key in the schema

In a scan with filter, i don't think there should be a difference in how we handle 3 vs 4

src/k2/dto/Expression.h
```diff
@@ -0,0 +1,106 @@
...
+    K2_PAYLOAD_FIELDS(fieldName, type, literal);
+};
+
+// An Expression in the expression model.
+// The operation is applied in the order that the children are in the vector. So a binary
+// operator like LT would be applied as valueChildren[0] < valueChildren[1]
+struct Expression {
+    Operation op = Operation::UNKNOWN;
+    std::vector<Expression> expressionChildren;
+    std::vector<Value> valueChildren;
```
jerryhfeng, line 56: Should we consider using SschematizeValue directly, i.e. Id + field Idx instead of fieldName in valueChildren of Expression? Since when the client construct the col/field reference Value, it has already resolved the column/field name with schema, so using fieldName still will be less efficient as it need to be resolved again.

src/k2/dto/Expression.cpp
iafuture: The api here matches along with the other question about what to do if a field isn't set.
If we believe that the correct model is to treat
- key is in schema but not set
- key is not in schema
the same, then this API has to return an empty optional in both cases, with nullLast set if non-null value was found

src/k2/dto/Expression.cpp
iafuture: Yes, in particular, i made it work off of .type field so that we could also consolidate with the SKV's DO_ON_NEXT_RECORD_FIELD macro

src/k2/dto/Expression.h
iafuture: The issue is that the schematized value is computed on each record separately on the server side. It has to be done this way for reference values. I was thinking of adding some optimization later on for literals, but since these need to be sent over the wire, the optimization has to be done again on the server side. I was thinking of some lazy caching for these literals.

src/k2/dto/FieldTypes.cpp
```diff
@@ -45,8 +45,14 @@ template <> FieldType TToFieldType<int16_t>() { return FieldType::INT16T; }
 template <> FieldType TToFieldType<int32_t>() { return FieldType::INT32T; }
 template <> FieldType TToFieldType<int64_t>() { return FieldType::INT64T; }
 template <> FieldType TToFieldType<float>() { return FieldType::FLOAT; }
+template <> FieldType TToFieldType<double>() { return FieldType::DOUBLE; }
+template <> FieldType TToFieldType<bool>() { return FieldType::BOOL; }
+template <> FieldType TToFieldType<FieldType>() { return FieldType::FIELD_TYPE; }
 
 // All conversion assume ascending ordering
+template <> String FieldToKeyString<bool>(const bool& field) {
+    return field? "1": "0";
```
jfunston, line 54: You need the type byte and terminator bytes here as well, or you can remove this function if you don't want to support bool fields as keys

src/k2/dto/FieldTypes.h
```diff
@@ -34,11 +34,15 @@ namespace dto {
...
-    INT64T = 4,
-    FLOAT = 5, // Not supported as key field for now
+    STRING, // NULL characters in string is OK
+    INT16T,
+    INT32T,
+    INT64T,
+    FLOAT, // Not supported as key field for now
+    DOUBLE,  // Not supported as key field for now
+    BOOL,
+    FIELD_TYPE,
```
jfunston, line 44: What is this one for?

iafuture: This is needed for determining the type of a reference field as part of an expression. We discussed this being the way users can perform schema migrations by discovering records where the desired column isn't upgraded

src/k2/dto/FieldTypes.cpp
iafuture: removed

src/k2/dto/SKVRecord.cpp
```diff
@@ -32,7 +32,7 @@ namespace dto {
 
 void SKVRecord::skipNext() {
     if (fieldCursor >= schema->fields.size()) {
-        throw SKVRecordException("Schema not followed in record serialization");
+        throw NoFieldFoundException();
```
jfunston, line 35: Why not have the string to give a little more info?

src/k2/dto/FieldTypes.h
jfunston: I still don't understand. I don't see it used anywhere unless I missed it. We can talk about it next week.

jfunston (changes requested):

src/k2/dto/Expression.cpp
```diff
@@ -0,0 +1,475 @@
...
+        throw InvalidExpressionException();
+    }
+    SchematizedValue aVal(valueChildren[0], rec);
+    SchematizedValue bVal(valueChildren[1], rec);
+    if (aVal.type != FieldType::STRING || bVal.type != FieldType::STRING) {
+        throw TypeMismatchException();
+    }
+    auto aOpt = std::get<1>(aVal.get<String>());
+    auto bOpt = std::get<1>(bVal.get<String>());
+    if (!bOpt) return true;   // all strings start with nothing
```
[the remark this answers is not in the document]
iafuture: fixed

src/k2/dto/Expression.h
```diff
@@ -0,0 +1,115 @@
...
+    bool isReference() const { return !fieldName.empty();}
+};
+
+// An Expression in the expression model.
+// The operation is applied in the order that the children are in the vector. E.g. a binary
+// operator like LT would be applied as valueChildren[0] < valueChildren[1]
+struct Expression {
+    Operation op = Operation::UNKNOWN;
+    std::vector<Value> valueChildren;
+    std::vector<Expression> expressionChildren;
```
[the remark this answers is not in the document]
iafuture: resolved

src/k2/dto/FieldTypes.h
```diff
@@ -41,20 +41,38 @@ struct SKVKeyEncodingException : public std::exception {
...
+template<> String FieldToKeyString<int64_t>(const int64_t&);
+template<> String FieldToKeyString<String>(const String&);
+
+// all other types are not supported as key fields
 template <typename T>
-String FieldToKeyString(const T& field);
+String FieldToKeyString(const T&) {
+    std::ostringstream msg;
+    msg << "Key encoding for " << TToFieldType<T>() << " not implemented yet";
+    throw std::runtime_error(msg.str());
```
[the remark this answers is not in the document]
iafuture: added

src/k2/dto/Expression.h
```diff
@@ -0,0 +1,115 @@
...
+    bool AND_handler(SKVRecord& rec);
+    bool OR_handler(SKVRecord& rec);
+    bool XOR_handler(SKVRecord& rec);
+    bool NOT_handler(SKVRecord& rec);
+};
+
+// helper builder: creates a value literal
+template <typename T>
+inline Value makeValueLiteral(T&& literal) {
+    Value result;
```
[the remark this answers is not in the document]
iafuture: good catch! fixed

src/k2/dto/SKVRecord.cpp
iafuture: added

jfunston (approved):

PR #69, opened 2020-10-21
Use SKV Partial Update in TPC-C Benchmark

YazhiFeng:


src/k2/cmd/tpcc/schema.h
```diff
@@ -56,6 +56,32 @@ seastar::future<k2::WriteResult> writeRow(ValueType& row, k2::K2TxnHandle& txn)
...
+            K2DEBUG("partialUpdateRow failed: " << result.status);
+            return seastar::make_exception_future<k2::PartialUpdateResult>(std::runtime_error("partialUpdateRow failed!"));
+        }
+
+        return seastar::make_ready_future<k2::PartialUpdateResult>(std::move(result));
+    });
+}
+
+template<typename ValueType>
+seastar::future<k2::PartialUpdateResult>
```
jerryhfeng, line 73: Suggestion: Could this fn combined into above one with input param vector element type parameterized?

jfunston: yeah, you can add a template parameter for the fieldsToUpdate type, and then you only need one function.

src/k2/cmd/tpcc/transactions.h
```diff
@@ -279,10 +284,13 @@ class NewOrderT : public TPCCTxn
                     *(line.Amount) = *(item.Price) * *(line.Quantity);
                     _total_amount += *(line.Amount);
                     line.DistInfo = stock.getDistInfo(*(line.DistrictID));
-                    updateStockRow(stock, line);
```
jfunston, line 282: You should delete the old definition of this function too.

jfunston (changes requested):

YazhiFeng:
Add a template parameter for partialUpdateRow and remove old func (updateStockRow).

jfunston (approved):

PR #71, opened 2020-10-21
SKV Code improvements

### LINKED ISSUE #1: Fix orphaned WIs for transactions with more than one write to the same key
If a transaction does more than one write to the same key, a write intent will be created for each one. The client will add the key multiple times to the write set, but only one write intent will be finalized.

### LINKED ISSUE #63: SKV code improvement

jfunston:
- Combine write and partial update into single RPC
- Fix exception throwing
- Fixes #1 
- Closes #63

### DIFF
--- a/src/k2/dto/ControlPlaneOracle.cpp
+++ b/src/k2/dto/ControlPlaneOracle.cpp
@@ -38,7 +38,7 @@ void Schema::setKeyFieldsByName(const std::vector<String>& keys, std::vector<uin
         }
 
         if (!found) {
-            throw std::runtime_error("Failed to find field by name");
+            throw CPOClientException("Failed to find field by name");
         }
     }
 }
--- a/src/k2/dto/ControlPlaneOracle.cpp
+++ b/src/k2/dto/ControlPlaneOracle.cpp
@@ -38,7 +38,7 @@ void Schema::setKeyFieldsByName(const std::vector<String>& keys, std::vector<uin
         }
 
         if (!found) {
-            throw new std::runtime_error("Failed to find field by name");
+            throw std::runtime_error("Failed to find field by name");
         }
     }
 }
--- a/src/k2/dto/FieldTypes.cpp
+++ b/src/k2/dto/FieldTypes.cpp
@@ -180,7 +180,7 @@ template <> String FieldToKeyString<int32_t>(const int32_t& field) {
 
 template <> String FieldToKeyString<float>(const float& field) {
     (void) field;
-    throw std::runtime_error("Key encoding for float is not implemented yet");
+    throw SKVKeyEncodingException("Key encoding for float is not implemented yet");
 }
 
 String NullFirstToKeyString() {
--- a/src/k2/dto/FieldTypes.cpp
+++ b/src/k2/dto/FieldTypes.cpp
@@ -180,7 +180,7 @@ template <> String FieldToKeyString<int32_t>(const int32_t& field) {
 
 template <> String FieldToKeyString<float>(const float& field) {
     (void) field;
-    throw new std::runtime_error("Key encoding for float is not implemented yet");
+    throw std::runtime_error("Key encoding for float is not implemented yet");
 }
 
 String NullFirstToKeyString() {
--- a/src/k2/dto/K23SI.h
+++ b/src/k2/dto/K23SI.h
@@ -207,41 +207,17 @@ struct K23SIWriteRequest {
     // use the name "key" so that we can use common routing from CPO client
     Key key; // the key for the write
     SKVRecord::Storage value; // the value of the write
-    K2_PAYLOAD_FIELDS(pvid, collectionName, mtr, trh, isDelete, designateTRH, key, value);
+    std::vector<uint32_t> fieldsToUpdate; // if size() > 0 then this is a partial update
+    K2_PAYLOAD_FIELDS(pvid, collectionName, mtr, trh, isDelete, designateTRH, key, value, fieldsToUpdate);
     friend std::ostream& operator<<(std::ostream& os, const K23SIWriteRequest& r) {
         return os << "{pvid=" << r.pvid << ", colName=" << r.collectionName
                   << ", mtr=" << r.mtr << ", trh=" << r.trh << ", key=" << r.key << ", isDelete="
-                  << r.isDelete << ", designate=" << r.designateTRH << "}";
+                  << r.isDelete << ", designate=" << r.designateTRH << ", fieldsToUpdate.size()=" 
+                  << r.fieldsToUpdate.size() << "}";
     }
 };
 
 struct K23SIWriteResponse {
-    K2_PAYLOAD_EMPTY;
-};
-
-struct K23SIPartialUpdateRequest{
-    Partition::PVID pvid; // the partition version ID. Should be coming from an up-to-date partition map
-    String collectionName; // the name of the collection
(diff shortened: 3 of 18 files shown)


src/k2/dto/FieldTypes.cpp
```diff
@@ -180,7 +180,7 @@ template <> String FieldToKeyString<int32_t>(const int32_t& field) {
 
 template <> String FieldToKeyString<float>(const float& field) {
     (void) field;
-    throw new std::runtime_error("Key encoding for float is not implemented yet");
+    throw std::runtime_error("Key encoding for float is not implemented yet");
```
iafuture, line 183: We should probably create a type for this exception and throw that out instead of generic runtime_error. Same goes for all the other exception changes in this PR i think

src/k2/dto/ControlPlaneOracle.cpp
```diff
@@ -38,7 +38,7 @@ void Schema::setKeyFieldsByName(const std::vector<String>& keys, std::vector<uin
         }
 
         if (!found) {
-            throw new std::runtime_error("Failed to find field by name");
+            throw std::runtime_error("Failed to find field by name");
```
iafuture, line 41: exc type instead?

src/k2/module/k23si/Module.cpp
```diff
@@ -756,7 +751,7 @@ bool K23SIPartitionModule::_makeFieldsForDiffVersion(dto::Schema& schema, dto::S
     return true;
 }
 
-bool K23SIPartitionModule::_parsePartialRecord(dto::K23SIPartialUpdateRequest& request, std::deque<dto::DataRecord>& versions) {
+bool K23SIPartitionModule::_parsePartialRecord(dto::K23SIWriteRequest& request, std::deque<dto::DataRecord>& versions) {
```
iafuture, line 754: I don't think we need to parse the entire deque here. Just a ref to the previous version should suffice

src/k2/module/k23si/Module.cpp
```diff
@@ -756,7 +751,7 @@ bool K23SIPartitionModule::_makeFieldsForDiffVersion(dto::Schema& schema, dto::S
...
     auto schemaIt = _schemas.find(request.key.schemaName);
     if (schemaIt == _schemas.end()) return false;
     auto schemaVer = schemaIt->second.find(request.value.schemaVersion);
```
iafuture, line 757: I think this schema validation should happen via helper in handleWrite() and we should return a status indicating that schema wasn't found

src/k2/module/k23si/Module.cpp
```diff
@@ -840,16 +834,9 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, dto::K23SI_M
         auto& rec = versions[0];
         auto& rqmtr = request.mtr;
     
-        // Clean up if req and WI are in the same transaction
```
iafuture, line 843: it is interesting that this has to be removed in order for partial updates to work correctly (p-update over WI from same txn scenario). We should probably have a test case for that if we don't already

iafuture (changes requested):

src/k2/module/k23si/Module.cpp
jfunston: Yes, handleWrite was already checking for the schema to exist, so I removed the checks here.

src/k2/module/k23si/Module.cpp
jfunston: I checked and the existing tests were all partial update over write intent, so I added one that is partial update over committed version.

jerryhfeng (changes requested):

src/k2/dto/K23SI.h
```diff
@@ -207,44 +207,20 @@ struct K23SIWriteRequest {
...
     SKVRecord::Storage value; // the value of the write
-    K2_PAYLOAD_FIELDS(pvid, collectionName, mtr, trh, isDelete, designateTRH, key, value);
+    std::vector<uint32_t> fieldsToUpdate; // if size() > 0 then this is a partial update
+    K2_PAYLOAD_FIELDS(pvid, collectionName, mtr, trh, isDelete, designateTRH, key, value, fieldsToUpdate);
     friend std::ostream& operator<<(std::ostream& os, const K23SIWriteRequest& r) {
         return os << "{pvid=" << r.pvid << ", colName=" << r.collectionName
                   << ", mtr=" << r.mtr << ", trh=" << r.trh << ", key=" << r.key << ", isDelete="
-                  << r.isDelete << ", designate=" << r.designateTRH << "}";
+                  << r.isDelete << ", designate=" << r.designateTRH << ", fieldsToUpdate.size()=" 
+                  << r.fieldsToUpdate.size() << "}";
```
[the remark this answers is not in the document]
jfunston: The size of fieldsToUpdate determines if it is a partial update or not. If size is 0 then it is a normal full write.

src/k2/module/k23si/Module.cpp
```diff
@@ -840,16 +839,9 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, dto::K23SI_M
...
-        if (rec.txnId.mtr == rqmtr) {
-            _queueWICleanup(std::move(versions[0]));
-            versions.pop_front();
```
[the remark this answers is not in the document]
jfunston: Yes, we want to remove the old WI, but we need to move it to later in the function after the full record is reconstructed for a partial update, then the WI is removed.

src/k2/module/k23si/Module.cpp
```diff
@@ -870,117 +862,32 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, dto::K23SI_M
...
-    if (!_parsePartialRecord(request, versions)) {
-        K2DEBUG("Partition: " << _partition << ", can not parse partial record for key " << request.key);
-        versions[0].value.fieldData.seek(0);
-        return RPCResponse(dto::K23SIStatus::BadParameter("missing fields or can not interpret partialUpdate"), dto::K23SIPartialUpdateResponse{});
-    }
-
     // Clean up if req and WI are in the same transaction
-    if (versions.size() > 0 && versions[0].status == dto::DataRecord::WriteIntent && request.mtr == versions[0].txnId.mtr) {
-        _queueWICleanup(std::move(versions[0]));
+    if (versions.size() > 0 && versions[0].txnId.mtr == request.mtr) {
```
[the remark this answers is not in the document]
jfunston: I think my answer to the previous question covers this but if not it may be helpful to look at the full function instead of the diff. The diff is hard to read because it is showing the combined handleWrite with some old code from the partial update.

src/k2/module/k23si/Module.cpp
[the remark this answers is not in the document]
jerryhfeng: Oh, missed that. So, we re-construct new WI from existing WI, I guess it is ok and makes more sense to re-construct directly from latest committed version.

src/k2/dto/K23SI.h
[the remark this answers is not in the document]
jerryhfeng: I know, just suggest it to be more explicitly with a flag or something to say this is a partial update. Trying to be kind to the developer who will read the trace and debug the system later:-)

iafuture (dismissed):

jerryhfeng (approved):

Issue #59, opened 2020-10-13
Implement reverse scan for query


jfunston:

jfunston:
Reverse scan is a bit more complicated than forward scan for the SKV query because of the routing of multi-partition requests. When the scan reaches a partition boundary, the server needs to specify a key to use as the continuation token. The continuation token key will be used by the PartitionRequest method to route the request to the server based on the partition map. The request also need to work correctly in the situation of a partition map change (such as partition split or merge). As the code is now we do not have a good way to do this for a reverse scan because the partition map interprets partition ends as exclusive (in other words, with a key alone we don't have a way to specify the "previous partition").


To support reverse scan we need to add an "exclusive" flag to the query response continuation token. This will tell the client to consider the token exclusive. The query request will also need an "exclusiveStart" flag that tells the server whether the start key should be included in the result set. PartitionRequest will also need to be modified with a new parameter that tells it whether to consider end keys as inclusive and start keys as exclusive in the partition map (it can have a default value of false so that current uses of PartitionRequest do not need to be changed). When the client gets a response back with the exclusive flag in the continuation token set to true, then it needs to set the flag in the PartitionRequest and the exclusiveStart flag in the query request.


An empty partition and range key string should mean "end of table" for a start key of a reverse scan and an empty partition and range key should mean "start of table" for an end key of a reverse scan. The user creates these with SKVRecords where no serialize methods are called. Note that this is different than if the user calls skipNext on their SKVRecords. If the user specifies "end of table" for the start key then the exclusive flags will need to be set for the first request.

jfunston:
Some background information on PartitionRequest for extra context. PartitionRequest wraps an RPC that is meant to be handled by a logical data partition (as opposed to a particular physical server). It uses the partition map and the key from the request to find the correct destination. If the partition map is stale, PartitionRequest automatically handles refreshing it. PartitionRequest also handles retries when the response status indicates a retriable error. We want any data-path requests to have this functionality so they should all use this interface.

YazhiFeng:
Implement reverse scan for query:
1. 'exclusiveToken' parameter is added to the QueryResponse. It indicate whether 'nextToScan' key is included in the next round of scan. 
2. 'exclusiveKey' parameter is added to the QueryRequest. It will be set according to the QueryResponse. For now, I do not add an api to set 'exclusiveKey' be true at the first time.
[...]

PR #72, opened 2020-10-23
Implement projection for Query

YazhiFeng:
Projection has been implemented and a test case is added.


src/k2/module/k23si/Module.cpp
```diff
@@ -302,8 +302,17 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
...
+                // TODO apply filter 
+                // apply projection if the user call addProjection
+                if (request.projection.size() == 0) {
+                    // want all fields 
+                    response.results.push_back(viter->value.share());
+                } else {
+                    // serialize partial SKVRecord according to projection
+                    dto::SKVRecord::Storage storage;
+                    _makeProjection(viter->value, request, storage);
+                    response.results.push_back(storage.share());
```
jfunston, line 314: You can use std::move(storage) instead of share()

src/k2/module/k23si/Module.cpp
```diff
@@ -784,6 +793,38 @@ bool K23SIPartitionModule::_parsePartialRecord(dto::K23SIPartialUpdateRequest& r
...
+            // write field value into payload
+            _copyPayloadBaseToUpdate(fullRec.fieldData, payload, schema.fields[i].type);
+            excludedFields[i] = false;
+        }
+    }
+    
+    // set cursor(0) of base payload
+    fullRec.fieldData.seek(0);
+
+    projectionRec.excludedFields = excludedFields;
```
jfunston, line 821: You can use std::move here too

test/k23si/QueryTest.cpp
```diff
@@ -120,14 +121,15 @@ class QueryTest {
 
 seastar::future<> doQuery(const k2::String& start, const k2::String& end, int32_t limit, bool reverse, 
                           uint32_t expectedRecords, uint32_t expectedPaginations, 
-                          k2::Status expectedStatus=k2::dto::K23SIStatus::OK) {
+                          k2::Status expectedStatus=k2::dto::K23SIStatus::OK,
+                          std::vector<k2::String> projection=std::vector<k2::String>()) {
```
jfunston, line 125: I think you should change this function so it returns the result set to the caller, then the tests for projection can test that the correct data is returned. The already existing tests can just ignore the result.

test/k23si/QueryTest.cpp
```diff
@@ -248,10 +273,18 @@ seastar::future<> runScenario01() {
     });
 }
 
-// Query conflict cases
+// Query projection cases
 seastar::future<> runScenario02() {
     K2INFO("runScenario02");
 
+    K2INFO("Projection reads for giving fields");
+    return doQuery("a", "c", -1, false, 2, 1, k2::dto::K23SIStatus::OK, {"partition"});
```
jfunston, line 281: Like I mentioned above, let's check that the data is correct. Also, can you add a test that tries to project a field that is not part of the schema?

jfunston (changes requested):

YazhiFeng:
Use std::move to return value, func doQuery() returns val_set, and add a testcase.

jfunston (approved):
Looks good, thanks. You can start on https://github.com/futurewei-cloud/chogori-platform/issues/59 next. As always feel free to ask me questions on the issue or in the open source meeting.

src/k2/module/k23si/Module.cpp
```diff
@@ -784,6 +793,38 @@ bool K23SIPartitionModule::_parsePartialRecord(dto::K23SIWriteRequest& request,
     return true;
 }
 
+void K23SIPartitionModule::_makeProjection(dto::SKVRecord::Storage& fullRec, dto::K23SIQueryRequest& request, 
+        dto::SKVRecord::Storage& projectionRec) {
+    auto schemaIt = _schemas.find(request.key.schemaName);
+    auto schemaVer = schemaIt->second.find(fullRec.schemaVersion);
+    dto::Schema& schema = schemaVer->second;
+    std::vector<bool> excludedFields(schema.fields.size(), true);   // excludedFields for projection
+    Payload payload(Payload::DefaultAllocator);                     // payload for projection
```
jerryhfeng, line 802: Minor: Could you rename it to projectedPayload for better readability.

PR #73, opened 2020-10-27
Change Module.cpp to use cast apply macro

### LINKED ISSUE #70: Reduce code duplication of switch/case statements on FieldTypes

jfunston:
Also add license to Expression
Closes #70

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -472,52 +472,11 @@ std::size_t K23SIPartitionModule::_findField(const dto::Schema schema, k2::Strin
     return fieldNumber;
 }
 
-bool K23SIPartitionModule::_advancePayloadPosition(Payload& payload, dto::FieldType type) {
-    switch (type) {
-    case k2::dto::FieldType::STRING : {
-        k2::String value;
-        bool success = payload.read(value);
-        if (!success) {
-            return false;
-        }
-        break;
-    }
-    case k2::dto::FieldType::INT16T : {
-        int16_t value;
-        bool success = payload.read(value);
-        if (!success) {
-            return false;
-        }
-        break;
-    }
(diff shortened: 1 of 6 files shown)


src/k2/module/k23si/Module.cpp
```diff
@@ -472,105 +472,45 @@ std::size_t K23SIPartitionModule::_findField(const dto::Schema schema, k2::Strin
...
-    }
-    default :
-        return false;
-    }// end switch
-    return true;
+template <typename T>
+void _advancePayloadPosition(const dto::SchemaField& field, Payload& payload, bool& success) {
+    (void) field;
+    T value{};
+    success = payload.read(value);
```
iafuture, line 479: I wonder if we should add a `skip<T>()` method to payload so that we don't have to copy bytes that aren't needed if we want to skip ahead

src/k2/module/k23si/Module.cpp
```diff
@@ -472,105 +472,45 @@ std::size_t K23SIPartitionModule::_findField(const dto::Schema schema, k2::Strin
...
-    case k2::dto::FieldType::FLOAT : {
-        float value;
-        bool success = base.read(value);
-        if (!success) {
-            return false;
-        }
-        update.write(value);
-        break;
+template <typename T>
+void _copyPayloadBaseToUpdate(const dto::SchemaField& field, Payload& base, Payload& update, bool& success) {
```
iafuture, line 483: i think we can also move this into Payload to avoid 1 extra copy

jfunston:
Both skip and copy would be useful in payload, but can we save that for a separate task?

iafuture:
> Both skip and copy would be useful in payload, but can we save that for a separate task?

definitely

iafuture (approved):

PR #77, opened 2020-10-30
Implement reverse scan for query

YazhiFeng:


jfunston:
Hi Yazhi, thanks for the PR, I will review soon. For your next tasks you can work on (feel free to do them in any order):
https://github.com/futurewei-cloud/chogori-platform/issues/78
https://github.com/futurewei-cloud/chogori-platform/issues/79
https://github.com/futurewei-cloud/chogori-platform/issues/80

src/k2/module/k23si/Module.cpp
```diff
@@ -233,12 +242,12 @@ bool K23SIPartitionModule::_isScanDone(const IndexerIterator& it, const dto::K23
 
 // Helper for handleQuery. Returns continuation token (aka response.nextToScan)
 dto::Key K23SIPartitionModule::_getContinuationToken(const IndexerIterator& it,
-                    const dto::K23SIQueryRequest& request, size_t response_size) {
+                    dto::K23SIQueryRequest& request, size_t response_size) {
```
jfunston, line 245: I think it is better to have this function take the dto::K23SIQueryResponse and fill in the token and flag, rather than change the flag in the request.

src/k2/module/k23si/Module.h
```diff
@@ -171,6 +179,32 @@ class K23SIPartitionModule {
         return dto::K23SIStatus::OK("");
     }
 
+    Status _validateQueryRequest(const dto::K23SIQueryRequest& request) const {
```
jfunston, line 182: You should be able to refactor this with validateReadRequest to reduce some duplication, since only the last to checks here are specific to queries.

src/k2/module/k23si/Module.cpp
```diff
@@ -256,34 +265,41 @@ dto::Key K23SIPartitionModule::_getContinuationToken(const IndexerIterator& it,
...
             return RPCResponse(dto::K23SIStatus::OperationNotAllowed("Query not implemented for hash partitioned collection"), dto::K23SIQueryResponse{});
     }
-    if (request.reverseDirection) {
-            return RPCResponse(dto::K23SIStatus::OperationNotAllowed("Reverse scan query not fully implemented"), dto::K23SIQueryResponse{});
-    }
 
-    IndexerIterator key_it = _initializeScan(request.key, request.reverseDirection);
+    IndexerIterator key_it = _initializeScan(request.key, request.reverseDirection, request.exclusiveKey);
 
     for (; !_isScanDone(key_it, request, response.results.size());
```
jfunston, line 304: If exclusiveKey is true in the request then the server should not return it in the result set

test/k23si/QueryTest.cpp
```diff
@@ -251,10 +252,46 @@ seastar::future<> runScenario01() {
     });
 }
 
-// Query projection cases
+// Simple reverse scan queries with no filter or projection
 seastar::future<> runScenario02() {
```
jfunston, line 256: A few points on the tests:
- Adding a test document like docs/RFC/K23SI_testing.md would be helpful here to make sure we cover all the corner cases
- I think at least one of the tests should check the order of the results to see if they are in fact in reverse order
- Let's add a TODO to add test cases when we are capable of changing the partition map

src/k2/dto/Collection.cpp
```diff
@@ -119,22 +119,32 @@ PartitionGetter::PartitionGetter(Collection&& col) : collection(std::move(col))
     }
 }
 
-PartitionGetter::PartitionWithEndpoint& PartitionGetter::getPartitionForKey(const Key& key) {
+PartitionGetter::PartitionWithEndpoint& PartitionGetter::getPartitionForKey(const Key& key, bool reverse, bool exclusiveKey) {
```
jfunston, line 122: I think we should have some unit tests for this function since it is getting more complicated and is fundamental to the operation of the system.

jfunston (changes requested):

test/dto/PartitionTest.cpp
```diff
@@ -0,0 +1,127 @@
...
+        k2::dto::PartitionGetter::PartitionWithEndpoint& part = _pgetter.getPartitionForKey(key);
+        K2INFO("startkey:" << part.partition->startKey << ", endkey:" << part.partition->endKey << ", endpoint:" 
+               << *(part.partition->endpoints.begin()) << ", TXEndpoint:" << (*part.preferredEndpoint).getURL());
+        //K2EXPECT(part.partition->startKey, "");
+    });
+}
+
+};
+
+int main(int argc, char** argv) {
```
jfunston, line 117: I think you could have done this as a unit test instead of an integration test, but you don't need to change it now if you don't want to.

jfunston:
Good work, I don't have any more comments after the partition getter tests are done. You will also need to rebase on master before I can merge it.

YazhiFeng:
Partition getter tests are done.

jfunston (approved):

docs/RFC/Partition_testing.md
```diff
@@ -0,0 +1,36 @@
...
+| using a null-empty key to get Partition with default reverse and exclusiveKey flag | partition that key belongs          |                     |
+| using an empty key to get Partition with default reverse and exclusiveKey flag | First partition                     |                     |
+| using a null-empty key to get Partition with reverse flag set to be TRUE and default exclusiveKey flag | partition that key belongs          |                     |
+| using an empty key to get Partition with reverse flag set to be TRUE and default exclusiveKey flag | last partition                      |                     |
+| using a null-empty key to get Partition with reverse and exclusiveKey flag set to be TRUE. the key is NOT the Start key of any partitions. | partition that key belongs          |                     |
+| using a null-empty key to get Partition with reverse and exclusiveKey flag set to be TRUE. the key is the Start key of a partition. | previous partition that key belongs |                     |
+| using an empty key to get Partition with reverse and exclusiveKey flag set to be TRUE. | last partition                      |                     |
```
jerryhfeng, line 36: Suggestion: This and next test design below should not be under RFC folder, as this folder is for design spec for proposed features. 

Even better, these seems more test code comments, instead of test case designs. So maybe put inside the code together as comments will be better, help to understand the code and easy to keep these description consistent with code when test case evolves.

src/k2/cpo/client/CPOClient.h
```diff
@@ -77,17 +77,18 @@ class CPOClient {
     // is not assigned or if there was a retryable error. It allows only one outstanding
     // request for a given collection.
     template <typename ClockT=Clock>
-    seastar::future<Status> GetAssignedPartitionWithRetry(Deadline<ClockT> deadline, const String& name, const dto::Key& key, uint8_t retries = 1) {
+    seastar::future<Status> GetAssignedPartitionWithRetry(Deadline<ClockT> deadline, const String& name, const dto::Key& key, 
+                                    bool reverse = false, bool excludedKey = false, uint8_t retries = 1) {
```
jerryhfeng, line 81: Minor: parameter naming conventions - 
1. Try to avoid generic variable name unless the scope/context is well defined - "name" will be much better to be collectionName as this function param here.  While in CollectionGetRequest, data member "name" is ok, since its context is clearly for a collection, and if choosing "collectionName" here will even be more redundant. 
2. In general, Boolean param should start with sth like "is...", "has...", "contains..." etc, unless it is subjective and obvious, here, "reverse" may be ok, excludedKey for sure is not clear as "isExcludedKey". 

Also, as "reverse" only make sense to be set to true when "isExcludedKey" is set to true. You may want to put excludedKey before reverse in the parameter list, and check when "reverse" is set, isExcludedKey must be true as well?

docs/RFC/Partition_testing.md
```diff
@@ -0,0 +1,36 @@
...
+### Test setup
+
+- start a cluster with three k2 nodes.
+- assign collection with range partition. Set the range of each partition as follows: ["", c), [c, e), [e, "")
+
+### Test cases
+
+| test case                                                    | Expected result                     | Possible fix needed |
+| ------------------------------------------------------------ | ----------------------------------- | ------------------- |
+| using a null-empty key to get Partition with default reverse and exclusiveKey flag | partition that key belongs          |                     |
```
jerryhfeng, line 30: May be adding the input explicitly here before description make it easer to read, since you gave the ranges example above.

e.g. the first test case can add ("d", false, false) before the the description.

jerryhfeng (commented):
Looks good. 

Some minor change suggestions, mostly for coding conventions.

PR #81, opened 2020-10-30
Add expression filter to SKV Query

### LINKED ISSUE #60: Implement filter for query

jfunston:
closes #60

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -288,12 +288,8 @@ std::tuple<Status, bool> K23SIPartitionModule::_doQueryFilter(dto::K23SIQueryReq
     try {
         keep = request.filterExpression.evaluate(record);
     }
-    catch(dto::NoFieldFoundException&) {
-        keep = false;
-    }
-    catch(dto::TypeMismatchException&) {
-        keep = false;
-    }
+    catch(dto::NoFieldFoundException&) {}
+    catch(dto::TypeMismatchException&) {}
     catch (dto::DeserializationError&) {
         status = dto::K23SIStatus::OperationNotAllowed("DeserializationError in query filter");
     }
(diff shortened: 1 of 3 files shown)


src/k2/module/k23si/Module.cpp
```diff
@@ -268,6 +268,42 @@ dto::Key K23SIPartitionModule::_getContinuationToken(const IndexerIterator& it,
     };
 }
 
+// Makes the SKVRecord and applies the request's filter to it. If Status is not OK,
```
iafuture, line 271: why does bad status means that it should be returned?

src/k2/module/k23si/Module.cpp
```diff
@@ -268,6 +268,42 @@ dto::Key K23SIPartitionModule::_getContinuationToken(const IndexerIterator& it,
...
+
+    dto::SKVRecord record(request.collectionName, versionIt->second);
+    record.storage = storage.share();
+    bool keep = false;
+    Status status = dto::K23SIStatus::OK("");
+
+    try {
+        keep = request.filterExpression.evaluate(record);
+    }
+    catch(dto::NoFieldFoundException&) {
```
iafuture, line 291: these two exceptions seem to be no-op. Why bother catching them?

src/k2/module/k23si/Module.cpp
jfunston: If the status is bad from doQueryFilter, it should be returned to the user in the query response. That is what I mean here, not sure if that answers your question.

src/k2/module/k23si/Module.cpp
jfunston: It needs to be caught so it doesn't propagate out. The keep = false is a no-op though and can be removed.

src/k2/module/k23si/Module.cpp
iafuture: Oh i see. it is the status that is returned. I thought you're suggesting that the record should be returned if the status out of the filter is not OK.

iafuture (approved):

PR #82, opened 2020-11-06
Add LogStream Support

piggesthjy:


src/k2/cpo/client/CPOClient.h
```diff
@@ -259,6 +260,61 @@ class CPOClient {
...
+    template<typename ClockT=Clock>
+    seastar::future<std::tuple<Status, dto::MetadataLogStreamRegisterResponse>> RegisterMetadataStreamLog(Deadline<ClockT> deadline, String name, String plogId) {
+        dto::MetadataLogStreamRegisterRequest request{.name = std::move(name), .plogId=std::move(plogId)};
+
+        Duration timeout = std::min(deadline.getRemaining(), cpo_request_timeout());
+        return RPC().callRPC<dto::MetadataLogStreamRegisterRequest, dto::MetadataLogStreamRegisterResponse>(dto::Verbs::CPO_METADATA_LOG_STREAM_REGISTER, request, *cpo, timeout)
+        .then([this, &request, deadline] (auto&& result) {
+            auto& [status, k2response] = result;
+
+            if (deadline.isOver()) {
```
jfunston, line 272: callRPC already does this for you, so you can remove this whole if block. Same comment for the rest of these functions.

src/k2/cpo/service/CPOService.cpp
```diff
@@ -491,4 +503,47 @@ CPOService::handlePersistenceClusterGet(dto::PersistenceClusterGetRequest&& requ
...
+
+seastar::future<std::tuple<Status, dto::MetadataLogStreamUpdateResponse>>
+CPOService::handleMetadataLogStreamUpdate(dto::MetadataLogStreamUpdateRequest&& request){
+    K2INFO("Received metadata log stream update request for " << request.name);
+    auto logstream = _metadataStreamLog.find(request.name);
+    if (logstream == _metadataStreamLog.end()) {
+        return RPCResponse(Statuses::S409_Conflict("metadata log does not exist"), dto::MetadataLogStreamUpdateResponse());
+    }
+
+    logstream->second[-1].offset = request.sealedOffset;
```
jfunston, line 530: Why is there a negative index here? If it is supposed to be that way, a comment explaining it would be good.

src/k2/module/k23si/Config.h
```diff
@@ -70,5 +70,11 @@ struct K23SIConfig {
 
     // the endpoint for the CPO
     ConfigVar<String> cpoEndpoint{"k23si_cpo_endpoint", "tcp+k2rpc://127.0.0.1:12345"};
+
+    // the name of the persistence cluster
+    ConfigVar<String> persistenceClusrerName{"k23si_persistence_cluster", "Persistence_Cluster_1"};
```
jfunston, line 75: Typo in variable name

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,287 @@
...
+    for (uint32_t i = 0; i < METADATA_PLOG_POOL_SIZE; ++i){
+        createFutures.push_back(_client.create());
+    }
+    return seastar::when_all_succeed(createFutures.begin(), createFutures.end())
+    .then([this] (auto&& responses){
+        // register the metadata plog info to CPO
+        for (auto& response: responses){
+            auto& [status, plogId] = response;
+            if (!status.is2xxOK()){
+                throw std::runtime_error("unable to create plog for Metadata LogStream");
```
jfunston, line 79: We are trying to use our own exceptions, instead of generic runtime_error, so you can define a LogStreamException or something like that and throw it instead. Also, should there be some cleanup of plogs that are already created?

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,287 @@
...
+}
+
+
+seastar::future<> 
+LogStream::write(Payload payload){
+    return _write(std::move(payload), 1);
+}
+
+seastar::future<> 
+LogStream::_write(Payload payload, uint32_t writeToWAL){
```
jfunston, line 214: writeToWAL would be better as a bool

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,287 @@
...
+        }
+        _plogInfo[writeToWAL].first = new_plogId;
+        _plogInfo[writeToWAL].second = 0;
+    }
+
+    uint32_t current_offset = _plogInfo[writeToWAL].second;
+    _plogInfo[writeToWAL].second += payload.getSize();
+
+    // write the incoming payload to current plog
+    return _client.append(_plogInfo[writeToWAL].first, current_offset,  std::move(payload))
```
jfunston, line 245: It seems like it is possible to leak plogs here, if we tried to allocate a new one from the pool above but the append call fails.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,287 @@
...
+            // Write the metadata to metadata log stream
+            if (writeToWAL){
+                Payload temp_payload([] { return Binary(4096); });
+                temp_payload.write(sealed_offest);
+                temp_payload.write(std::move(new_plogId));
+                return _write(std::move(temp_payload), 0);
+            }
+            // Write the metadata to CPO
+            else{
+                return _cpo.UpdateMetadataStreamLog(Deadline<>(_cpo_timeout()), _logStreamName, sealed_offest, std::move(new_plogId))
```
jfunston, line 274: So the CPO keeps track of every metadata plog? Is that what we want, or would it be better for the CPO to only store the root (first) metadata plog?

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,287 @@
...
+            K2INFO("unable to append the plog");
+            K2INFO(status);
+            throw std::runtime_error("unable to append the plog");
+        }
+        // If sealed_offest == 0, then it means after the incoming writing operation, the current plog is not excceed the size limit
+        if (sealed_offest == 0){
+            return seastar::make_ready_future<>();
+        }
+        // It the previous plog excceed the size limit, we will seal the previous plog and write the metadata to corrsponding place (Metadata Log Stream or CPO)
+        return _client.seal(sealed_plogId, sealed_offest)
```
jfunston, line 258: I think it is better to seal first before doing the append.

src/k2/persistence/logStream/LogStream.h
```diff
@@ -0,0 +1,85 @@
...
+    // How many metadata plogs it will create in advance
+    const static uint32_t METADATA_PLOG_POOL_SIZE = 4;
+
+    CPOClient _cpo;
+    PlogClient _client;
+    String _logStreamName;
+
+    std::vector<String> _walPlogPool; // The vector to store the created WAL plog Id
+    std::vector<String> _metadataPlogPool; // the vector to store the created Metadata plog Id
+    std::vector<std::pair<String, uint32_t>> _plogInfo; // _plogInfo[0] means the current metadata Plog Id and offset it holds, while _plogInfo[1] means the current WAL Plog Id and offset it holds,
```
jfunston, line 71: This would be better as two different variables. If you are worried about the _write implemenation, you can do something like this: "std::pair<String, unit32_t>& targetPlogInfo = writeToWAL ? _wallInfo : _metaInfo"

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,287 @@
...
+LogStream::read(String logStreamName){
+    K2INFO("Received Logstream Read Request for " << logStreamName);
+    // read the metadata of metadata log stream from CPO
+    return _cpo.GetMetadataStreamLog(Deadline<>(_cpo_timeout()), logStreamName)
+    .then([&] (auto&& response){
+        auto& [status, resp] = response;
+        if (!status.is2xxOK()) {
+            throw std::runtime_error("unable to obtain metadata log stream");
+        }
+        _streamLog = std::move(resp.streamLog);
```
jfunston, line 129: Using a class variable here means it won't be safe for concurrent calls to read, it would be better to have it allocated on the heap.

src/k2/persistence/logStream/LogStream_backup.cpp
```diff
@@ -0,0 +1,259 @@
+/*
```
jfunston, line 1: Accidentally added this file to the commit?

src/k2/persistence/plog_service/PlogServer.h
```diff
@@ -39,24 +39,26 @@ Copyright(c) 2020 Futurewei Cloud
...
-    Payload payload; 
-};
+
 
 class PlogServer
 {
 private:
     // the maximum size of each plog
-    const uint32_t PLOG_MAX_SIZE = 2 * 1024 * 1024;
+    const static uint32_t PLOG_MAX_SIZE = 16 * 1024 * 1024;
```
jfunston, line 48: very minor point, but you can use constexpr here

test/plog/LogStreamTest.cpp
```diff
@@ -0,0 +1,183 @@
...
+            for (auto& payload:payloads){
+                payload.seek(0);
+                payload.read(str);
+                K2EXPECT(str, header);
+            }
+            return seastar::make_ready_future<>();
+        });
+    }
+
+    seastar::future<> runTest4() {
```
jfunston, line 142: How many wal plogs does this use? How many metadata plogs? Does it exhaust the wal plog pool?

jfunston (changes requested):

src/k2/cpo/client/CPOClient.h
piggesthjy: fixed

src/k2/cpo/service/CPOService.cpp
piggesthjy: There is a bug before. Fixed

src/k2/module/k23si/Config.h
piggesthjy: fixed

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: added the own expection

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: fixed

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: Yes. The log stream will not communicate with the CPO frequently. If the Plog size = 1G, then we will update the CPO once if we have sealed 10M WAL Plogs

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: I am not sure whether I need to handle the failure case here. If we want to handle this case, the only thing we can do is that, when we detected an error, we should seal all the opened Plogs?

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: If we do the seal first, we need to wait for this seal to be done before we execute the next writing. I think this seems to be a lock, which we would like to avoid. And do the append first will not affect the correctness of this scenario, and can increase the throughput.

src/k2/persistence/logStream/LogStream.h
piggesthjy: fixed

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: fixed

src/k2/persistence/logStream/LogStream_backup.cpp
piggesthjy: removed

src/k2/persistence/plog_service/PlogServer.h
piggesthjy: fixed

test/plog/LogStreamTest.cpp
piggesthjy: For the current test, it will use 3 WAL plogs. Now I set the size of WAL plog pool to be 1, and create new plogs if the current WAL plog pool is empty. But currently, this test just uses 1 metadata plog. It is difficult for me to test the scenario that we have used more than 1 metadata plogs. I reduce the size of each plog from 16M to 100Bytes by my own to test this scenario. But in our code, we did not support changing the size of each plog. So I cannot implement this test. Do you think we should add this to our code as well?

src/k2/persistence/logStream/LogStream.cpp
jfunston: Maybe, I think you need to at least write to the metadata log first then, maybe Ivan or Jerry can help clarify.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,311 @@
...
+                throw k2::dto::PlogCreateError("unable to create plog for WAL");
+            }
+            _walPlogPool.push_back(std::move(plogId));
+        }
+        String plogId = _walPlogPool[_walPlogPool.size()-1];
+        _walPlogPool.pop_back();
+        _walInfo = std::move(std::make_pair(plogId, 0));
+
+        // write the metadata of the first WAL plog to metadata log streams
+        Payload temp_payload([] { return Binary(4096); });
```
jfunston, line 113: You can use Payload::DefaultAllocator() here

test/plog/LogStreamTest.cpp
jfunston: Maybe you can add a very small plog size for testing purposes?

src/k2/persistence/logStream/LogStream.cpp
jfunston: You don't need to fix it now but it would be good to add a TODO section to the design document. Another TODO is handling the case when the pool is empty which I see is an exception right now.

test/plog/LogStreamTest.cpp
piggesthjy: In our code, we use `constexpr static uint32_t PLOG_MAX_SIZE = 16 * 1024 * 1024;` to set the size of each plog. Do you think we should change this to be a variable and allow our plog client to set this value in the plog server?

jfunston: I see, you can leave it how it is, it isn't a big deal.

piggesthjy: leave a TODO here

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: fixed

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: If there is an ongoing seal request, I will first send the upcoming write request to the plog servers. However, if this write operation finished, it will not return success to the client until the ongoing seal request finished.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+seastar::future<>
+LogStream::create(String logStreamName){
+    if (_logStreamCreate){
+        K2INFO("Log streaam already created");
+        throw k2::dto::LogStreamExistError("Log stream already created");
+    }
+
+    _logStreamName = std::move(logStreamName);
+    _logStreamCreate = true;
```
jerryhfeng, line 68: I think you can only set _logStreamCreate to true at the last step inside this function, otherwise, there maybe client try to write into this LogStream even before you create any meta plogs, or wal plogs below.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+            _walPlogPool.push_back(std::move(plogId));
+        }
+        String plogId = _walPlogPool[_walPlogPool.size()-1];
+        _walPlogPool.pop_back();
+        _walInfo = std::move(std::make_pair(plogId, 0));
+
+        // write the metadata of the first WAL plog to metadata log streams
+        Payload temp_payload(Payload::DefaultAllocator);
+        temp_payload.write(std::move(plogId));
+        return _write(std::move(temp_payload), 0);
```
jerryhfeng, line 117: second parameter better use "false" instead of 0.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+seastar::future<> 
+LogStream::_switchPlog(Payload payload, bool writeToWAL){
+    std::pair<String, uint32_t>& targetPlogInfo = writeToWAL ? _walInfo : _metaInfo;
+
+    String new_plogId;
+    String sealed_plogId = targetPlogInfo.first;
+    uint32_t sealed_offest = targetPlogInfo.second;
+    
+    // change to the next available plog id
+    if (!writeToWAL){
```
jerryhfeng, line 270: Mingle WAL and meta data write into the same function is not a good idea due to coupling. No need to change it now, but we should naturally decouple them when we split wal and meta data from logstream.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+    targetPlogInfo.second += payload.getSize();
+
+    return _client.append(targetPlogInfo.first, current_offset, std::move(payload))
+    .then([&, writeToWAL, sealed_plogId, new_plogId, sealed_offest] (auto&& response){
+        auto& [status, resp] = response;
+        if (!status.is2xxOK()) {
+            throw k2::dto::PlogAppendError("unable to append the plog");
+        }
+        // If sealed_offest == 0, then it means after the incoming writing operation, the current plog is not excceed the size limit
+        return _client.seal(sealed_plogId, sealed_offest)
```
jerryhfeng, line 298: why seal after append new data, we can do it in parallel, right?

Also, what happens while the seal is ongoing, and append comes in and will it try to seal again (which is wrong)?

jerryhfeng (changes requested):

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+        });
+        
+    });
+}
+
+seastar::future<std::vector<Payload> >
+LogStream::_readContent(std::vector<Payload> payloads){
+    std::vector<dto::MetadataElement> streamLog;
+
+    //Format: plogId1, offset1, plogId2, offset2...
```
iafuture: if this is the expected format for elements in the log, you should create a struct for these elements and read/write from the payload in terms of structs, not individual fields. In fact i think your MetadataElement is exactly that

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+                return _readContent(std::move(payloads));
+            });
+        });
+        
+    });
+}
+
+seastar::future<std::vector<Payload> >
+LogStream::_readContent(std::vector<Payload> payloads){
+    std::vector<dto::MetadataElement> streamLog;
```
iafuture: what is a streamLog? I can't tell if the rest of the code is quite OK since I don't understand the algorithm here

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+                dto::MetadataElement log_entry{.name=pending_plogId, .offset=offset};
+                streamLog.push_back(std::move(log_entry));
+            }
+        } 
+        payload.clear();
+    }
+    dto::MetadataElement log_entry{.name=pending_plogId, .offset=0};
+    streamLog.push_back(std::move(log_entry));
+
+    return seastar::do_with(std::move(streamLog), [this] (auto& streamLog){
```
iafuture: what is this section of the code supposed to achieve?

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+        return _client.info(streamLog[streamLog.size()-1].name)
```
iafuture: this seems to be an API that does a remote call. I'd prefer if it has a better name to describe what it does (something like getPLOGInfo())

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+            }
+            streamLog[streamLog.size()-1].offset = std::get<0>(resp);
+            std::vector<seastar::future<std::tuple<Status, Payload> > > readFutures;
+            for (auto& log_entry: streamLog){
+                readFutures.push_back(_client.read(log_entry.name, 0, log_entry.offset));
+            }
+            return seastar::when_all_succeed(readFutures.begin(), readFutures.end());
+        })
+        .then([this] (std::vector<std::tuple<Status, Payload> >&& responses){
+            std::vector<Payload> newPayloads;
```
iafuture: the name newPayloads does not tell me at all what these are supposed to represent. It is the same as say "int32_t newInt32_t". It seems you're storing plogs in them, so a name that describes what plogs these are would be better

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+                newPayloads.push_back(std::move(resp));
+            }
+            return seastar::make_ready_future<std::vector<Payload> >(std::move(newPayloads));
+        });
+    });
+}
+
+
+seastar::future<> 
+LogStream::write(Payload payload){
```
iafuture: I think this api can be in terms of rvalue& instead of moved values;

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+    return _write(std::move(payload), true);
+}
+
+seastar::future<> 
+LogStream::_write(Payload payload, bool writeToWAL){
```
iafuture: rvalue& for the write payload?

Also this WAL makes no sense to me in the context of a LogStream

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+    if (!_logStreamCreate){
+        K2INFO("Log stream does not create");
```
iafuture: no need to log here i think. just throw .

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+    std::pair<String, uint32_t>& targetPlogInfo = writeToWAL ? _walInfo : _metaInfo;
+
+    if (targetPlogInfo.second + payload.getSize() < PLOG_MAX_SIZE){
+        return _client.append(targetPlogInfo.first, targetPlogInfo.second, std::move(payload)).
+        then([this, writeToWAL] (auto&& response){
+            auto& [status, resp] = response;
+            if (!status.is2xxOK()) {
+                throw k2::dto::PlogAppendError("unable to append the plog");
+            }
+            // Check weather there is a sealed request that did not receive the response
```
iafuture: typo: weather -> whether

src/k2/persistence/logStream/LogStream.h
```diff
@@ -0,0 +1,85 @@
...
+#include <k2/config/Config.h>
+#include <k2/cpo/client/CPOClient.h>
+#include <k2/transport/BaseTypes.h>
+#include <k2/transport/TXEndpoint.h>
+#include <k2/persistence/plog_client/PlogClient.h>
+
+
+namespace k2 {
+
+class LogStream{
```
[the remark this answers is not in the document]
piggesthjy: renamed

src/k2/persistence/plog_service/PlogServer.h
```diff
@@ -39,24 +39,26 @@ Copyright(c) 2020 Futurewei Cloud
...
+
+    // each PlogPage is a plog. It uses payload to store the data, and contains the sealed and offest as metadata
+    struct PlogPage {
```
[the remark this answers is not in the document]
piggesthjy: renamed

docs/PERSISTENCE.md
```diff
@@ -1,11 +1,12 @@
 [-UP-](./README.md)
 
+
 # Plog
-The persistence service in Chogori Project is called Plog. Plog will store all the data in Non-Volatile Memory(NVM), which provides similar performance comparing to the Dynamic random-access memory(DRAM), but can retrieve stored information even after having been power cycled. Since comparing to the DRAM, NVM has a higher random access latency. In our current design, this Plog is an append-only struct. The only operation that allowed to change the Plog is to append new data after current blocks. 
+The data model provided by Chogori Project is called Plog. In our current design, this Plog is an append-only struct. The only operation that allowed to change the Plog is to append new data after current blocks.
```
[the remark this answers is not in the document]
piggesthjy: changed

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -228,38 +228,63 @@ LogStream::_write(Payload payload, bool writeToWAL){
...
+                });
+            }
+            else{
+                return seastar::make_ready_future<>();
+            }
+        });
+    }
+    else{
+        // switch to a new plog
+        return _switchPlog(std::move(payload), writeToWAL);
```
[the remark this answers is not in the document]
piggesthjy: renamed

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -0,0 +1,343 @@
...
+
+    if (targetPlogInfo.second + payload.getSize() < PLOG_MAX_SIZE){
+        return _client.append(targetPlogInfo.first, targetPlogInfo.second, std::move(payload)).
+        then([this, writeToWAL] (auto&& response){
+            auto& [status, resp] = response;
+            if (!status.is2xxOK()) {
+                throw k2::dto::PlogAppendError("unable to append the plog");
+            }
+            // Check weather there is a sealed request that did not receive the response
+            if (_switched && writeToWAL){
```
iafuture: this flag _switched seems to indicate if there is a seal request in progress. A name such as _sealRequestInProgress would be better.

src/k2/persistence/logStream/LogStream.cpp
```diff
@@ -228,38 +228,63 @@ LogStream::_write(Payload payload, bool writeToWAL){
...
-                throw k2::dto::LogStreamBackupPlogError("no available backup wal plog");
-            }
-            new_plogId = _walPlogPool[_walPlogPool.size()-1];
-            _walPlogPool.pop_back();
+        new_plogId = _metadataPlogPool[_metadataPlogPool.size()-1];
+        _metadataPlogPool.pop_back();
+    }
+    else{
+        if (_walPlogPool.size() == 0){
+            throw k2::dto::LogStreamBackupPlogError("no available backup wal plog");
```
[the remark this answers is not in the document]
piggesthjy: Added the comments.
Replenish the pool is in _switchPlogAndWrite. But it can be implemented in parallel, which is not the current implementation

src/k2/persistence/plog_service/PlogServer.cpp
```diff
@@ -161,5 +164,16 @@ PlogServer::_handleSeal(dto::PlogSealRequest&& request){
     return RPCResponse(Statuses::S200_OK("sealed success"), std::move(response));
 };
 
+seastar::future<std::tuple<Status, dto::PlogInfoResponse>>
+PlogServer::_handleInfo(dto::PlogInfoRequest&& request){
```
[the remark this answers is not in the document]
piggesthjy: renamed

src/k2/persistence/logStream/LogStream.h
```diff
@@ -0,0 +1,90 @@
...
+    seastar::future<> init(String cpo_url, String persistenceClusrerName);
+
+    // create a log stream with a given name
+    seastar::future<> create(String logStreamName);
+
+    // write data to the log stream
+    seastar::future<> write(Payload payload);
+
+    // read all the data from this log stream
+    seastar::future<std::vector<Payload> > read(String logStreamName);
```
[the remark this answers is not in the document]
piggesthjy: Renamed and list a TODO here

src/k2/persistence/plog_client/PlogClient.h
```diff
@@ -63,6 +71,9 @@ class PlogClient {
     // seal a payload
     seastar::future<std::tuple<Status, uint32_t>> seal(String plogId, uint32_t offset);
 
+    // obtain the current offset and status of a plog
+    seastar::future<std::tuple<Status, std::tuple<uint32_t, bool>>> info(String plogId);
```
iafuture: getPLOGInfo is a better name i think

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: Set the _logStreamCreate=true at the penultimate step. The last step of the creation is to write the WAL Plog id to metadata Plog, which requires the _logStreamCreate==true

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: fixed

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: Listed a TODO here

src/k2/persistence/plog_service/PlogServer.h
```diff
@@ -76,6 +78,10 @@ class PlogServer
     //handle the seal request
     seastar::future<std::tuple<Status, dto::PlogSealResponse>>
     _handleSeal(dto::PlogSealRequest&& request);
+
+    //handle the seal request
```
iafuture, line 82: did you mean handle info request?

iafuture (changes requested):

src/k2/persistence/logStream/LogStream.cpp
piggesthjy: Yes, I agree. We can do them as well as replenish the Plog pool in parallel.

 Seal the same Plog twice will not happen. If we find the old Plog needs to be sealed, we will switch to the backup Plog from the Plog pool immediately. Then when the new append comes, it will write to the new Plog instead of the one that needs to be sealed.

PR #86, opened 2020-11-17
Add dto::Key interface to SKV client

jfunston:
This is to support rowID for SQL and allow more efficient operations when the key can be reused to avoid extra serialization.

### DIFF
--- a/src/k2/module/k23si/client/k23si_client.cpp
+++ b/src/k2/module/k23si/client/k23si_client.cpp
@@ -79,12 +79,12 @@ void K2TxnHandle::makeHeartbeatTimer() {
 
 std::unique_ptr<dto::K23SIReadRequest> K2TxnHandle::makeReadRequest(
                         const dto::Key& key, const String& collection) const {
-    return std::make_unique<dto::K23SIReadRequest>(dto::K23SIReadRequest{
+    return std::make_unique<dto::K23SIReadRequest>(
         dto::Partition::PVID(), // Will be filled in by PartitionRequest
         collection,
         _mtr,
         key
-    });
+    );
 }
 
 template <>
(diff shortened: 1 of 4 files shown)


src/k2/module/k23si/client/k23si_client.cpp
```diff
@@ -77,7 +77,18 @@ void K2TxnHandle::makeHeartbeatTimer() {
     });
 }
 
-std::unique_ptr<dto::K23SIReadRequest> K2TxnHandle::makeReadRequest(const dto::SKVRecord& record) const {
+std::unique_ptr<dto::K23SIReadRequest> K2TxnHandle::makeReadRequest(
+                        const dto::Key& key, const String& collection) const {
+    return std::make_unique<dto::K23SIReadRequest>(dto::K23SIReadRequest{
```
iafuture, line 82: you can avoid an extra move of the read request by using the emplace version here instead by just passing the ctor args to make_unique()

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -385,4 +385,9 @@ class K2TxnHandle {
     String _trh_collection;
 };
 
+// Normal use-case read interface, where the key fields of the user's SKVRecord are 
+// serialized and converted into a dto::Key
+template <>
```
iafuture, line 390: is this needed? "template <>"

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -285,7 +274,8 @@ class K2TxnHandle {
     }
 
     template <typename T1>
-    seastar::future<PartialUpdateResult> partialUpdate(T1& record, std::vector<k2::String> fieldsName) {
```
iafuture, line 288: I think it'd be better if we just gave overloads which take an extra dto::key. Then have the non-key API call into the key one.  same goes for the other changed APIs (make*, read, write)

iafuture (changes requested):

iafuture (approved):

PR #89, opened 2020-12-03
During TSO client initialization, when there are app requests (for ti…

jerryhfeng:
…mestamp) before TSO server end point is discoverd, let's these request wait after discovery is done.

### DIFF
--- a/src/k2/cmd/demo/tso_sample_app.cpp
+++ b/src/k2/cmd/demo/tso_sample_app.cpp
@@ -141,7 +141,7 @@ int main(int argc, char **argv)
     app.addOptions()
     ("tso_endpoint", bpo::value<k2::String>()->default_value("tcp+k2rpc://127.0.0.1:9000"), "URL of Timestamp Oracle (TSO), e.g. 'tcp+k2rpc://192.168.1.2:12345'");
 
-    app.addApplet<k2::TSO_ClientLib>(1s);
+    app.addApplet<k2::TSO_ClientLib>();
     app.addApplet<k2::TSOService>();
     app.addApplet<k2::sampleTSOApp>();
 
--- a/src/k2/tso/client/tso_clientlib.cpp
+++ b/src/k2/tso/client/tso_clientlib.cpp
@@ -174,7 +174,7 @@ seastar::future<Timestamp> TSO_ClientLib::GetTimestampFromTSO(const TimePoint& r
     {
         // if not ready to serve yet, wait on a new ready promise then call get this function self, as each promise can only chain one then lamda
         K2INFO("TSO Timestamp requested when not ready to serve, request pending...");
-        _promiseReadyToServe.emplace_back(seastar::promise<>());
+        _promiseReadyToServe.emplace_back();
         return _promiseReadyToServe.back().get_future()
             .then([this, triggeredTime = requestLocalTime] { return GetTimestampFromTSO(triggeredTime); });
     }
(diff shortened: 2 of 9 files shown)


src/k2/tso/client/tso_clientlib.h
```diff
@@ -73,6 +71,10 @@ class TSO_ClientLib
 
     bool _stopped{false};
 
+    // a promise/signal for ready to serve request when they come earlier than TSO server end point set up
+    volatile bool _readyToServe {false};
```
jfunston, line 75: No need for volatile since it is still single-threaded

src/k2/cmd/demo/tso_sample_app.cpp
```diff
@@ -141,7 +141,7 @@ int main(int argc, char **argv)
     app.addOptions()
     ("tso_endpoint", bpo::value<k2::String>()->default_value("tcp+k2rpc://127.0.0.1:9000"), "URL of Timestamp Oracle (TSO), e.g. 'tcp+k2rpc://192.168.1.2:12345'");
 
-    app.addApplet<k2::TSO_ClientLib>(1s);
+    app.addApplet<k2::TSO_ClientLib>();
```
jfunston, line 144: I don't think this problem is fixed is it? Here the problem is the client tried to connect the server before the server was ready, but this fix is for if the user of the client tries to get a timestamp before the client is ready.

src/k2/tso/client/tso_clientlib.h
jerryhfeng: Maybe a little bit paranoid here, but the  readyPromise.set_value(); could be executed before this line otherwise and could cause this task be schedule out while another client request may come in and wrong adding another promise and wait on it.

src/k2/cmd/demo/tso_sample_app.cpp
jerryhfeng: The true bug is that client request should not go through and wait until the server discover is done, which is this change about. This delay start hack was used to avoid one of two reasons this case could happen, which is that the server is not started yet. It is no longer needed here that we have delayed retry to connect to server in this tsoclient, which should be sufficient time wise in normal case, including in test.
So this hack should not be needed with the fix here.

src/k2/tso/client/tso_clientlib.cpp
```diff
@@ -159,6 +169,17 @@ seastar::future<Timestamp> TSO_ClientLib::GetTimestampFromTSO(const TimePoint& r
         return seastar::make_exception_future<Timestamp>(TSOClientLibShutdownException());
     }
 
+    // TSO client may not yet ready (discover the tso server endpoint), let the request wait in this case.
+    if (!_readyToServe)
+    {
+        // if not ready to serve yet, wait on a new ready promise then call get this function self, as each promise can only chain one then lamda
+        K2INFO("TSO Timestamp requested when not ready to serve, request pending...");
+        _promiseReadyToServe.emplace_back(seastar::promise<>());
```
iafuture, line 177: just call emplace_back() here. no need to supply a temporary promise to emplace_back

iafuture (changes requested):

src/k2/tso/client/tso_clientlib.cpp
jerryhfeng: done.

src/k2/tso/client/tso_clientlib.h
jerryhfeng: removed unnecessary volatile.

jfunston (approved):

iafuture (approved):

PR #91, opened 2020-12-19
Expose createCollection in k23si

jerryhfeng:


src/k2/module/k23si/client/k23si_client.h
```diff
@@ -125,7 +125,10 @@ class K23SIClient {
 
     seastar::future<> start();
     seastar::future<> gracefulStop();
+    // First version of makeCollection is deprecated, do not use.
+    // TODO: remove first version of makeCollection and update the test code.
```
jfunston, line 129: I don't think we need to deprecate the old makeCollection, it avoids a lot of boilerplate we don't need in a lot of use cases.

johnfangAFW (approved):
LGTM

PR #92, opened 2020-12-21
Api server

jfunston:
Add JSON HTTP API server that can be used as an applet. Can be used to register normal K2 RPCs for use over HTTP. Meant for administrative and debugging purposes. More work needed to be done to add the remaining inspect RPCs and add to the python client.


src/k2/infrastructure/APIServer.h
```diff
@@ -0,0 +1,104 @@
...
+#include <seastar/http/httpd.hh>
+
+namespace k2 {
+
+// Helper class for registering HTTP routes
+class api_route_handler : public seastar::httpd::handler_base  {
+public:
+    api_route_handler(std::function<seastar::future<String>(const std::string&)> h) : _handler(std::move(h)) {}
+
+    seastar::future<std::unique_ptr<seastar::httpd::reply>> handle(const String& path,
```
iafuture, line 40: in this reply, how do we set the status code? Should probably be always 200 OK

iafuture (changes requested):

iafuture (approved):

PR #100, opened 2021-01-11
SkV getKey fix for returned records

jfunston:
Also cleanup and tighten SKV interface

### DIFF
--- a/src/k2/dto/SKVRecord.cpp
+++ b/src/k2/dto/SKVRecord.cpp
@@ -63,8 +63,12 @@ void NoOp(std::optional<T> value, const String& fieldName, int n) {
     (void) n;
 };
 
+uint32_t SKVRecord::getFieldCursor() const {
+    return fieldCursor;
+}
+
 void SKVRecord::seekField(uint32_t fieldIndex) {
-    if (fieldIndex >= schema->fields.size()) {
+    if (fieldIndex > schema->fields.size()) {
         throw NoFieldFoundException();
     }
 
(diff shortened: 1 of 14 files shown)


jerryhfeng (dismissed):
Looks good to me.

Conditional approval regarding one of comments, assuming it is correct the have the havingKey/KeyContructed bool set in constructor is correct.

src/k2/dto/SKVRecord.cpp
```diff
@@ -93,15 +97,69 @@ Payload SKVRecord::getSharedPayload() {
...
+    }
+}
+
+void SKVRecord::constructKeyStrings() {
+    partitionKeys.resize(schema->partitionKeyFields.size());
+    rangeKeys.resize(schema->rangeKeyFields.size());
+
+    uint32_t beginCursor = getFieldCursor();
+    seekField(0);
+    for (size_t i = 0; i < partitionKeys.size() + rangeKeys.size(); ++i) {
```
iafuture, line 147: Are we preventing the same field from being used in both partition and range key? i.e. is it possible for pk.size() + rk.size() to be < allFields.size()

src/k2/dto/SKVRecord.cpp
```diff
@@ -93,15 +97,69 @@ Payload SKVRecord::getSharedPayload() {
...
+        DO_ON_NEXT_RECORD_FIELD(*this, makeKeyString, i);
```
iafuture, line 148: why pass this int if it isn't used by the func?

iafuture (changes requested):

iafuture (approved):

PR #103, opened 2021-01-18
Log system rewrite

iafuture:


src/k2/cmd/tpcc/Log.h
```diff
@@ -0,0 +1,29 @@
+/*
+MIT License
+
+Copyright(c) 2020 Futurewei Cloud
```
jfunston, line 4: 2021? Same thing for the other new files.

src/k2/cmd/tpcc/schema.h
```diff
@@ -32,41 +32,42 @@ Copyright(c) 2020 Futurewei Cloud
 #include <k2/transport/PayloadSerialization.h>
 
 #include "tpcc_rand.h"
-
-static const k2::String tpccCollectionName = "TPCC";
+#include "Log.h"
+using namespace k2;
```
jfunston, line 36: Is this needed to make the macros work or is it only to reduce some verbosity?

src/k2/common/FormattingUtils.h
```diff
@@ -0,0 +1,130 @@
...
+    friend void from_json(const nlohmann::json& j, CType& o) {                                    \
+        _K2_FROM_JSON(__VA_ARGS__)                                                                \
+    }
+
+// Generating enums with formatting. the _IC version is to be used when inside a class.
+// e.g.
+// namespace k2::dto {
+// K2_DEF_ENUM(TxnStatus, Created, Aborted, Committed) )
+// }
+#define K2_DEF_ENUM_IC(Name, ...)                                                   \
```
jfunston, line 72: Probably should use something more specific than "Name" to prevent aliasing

src/k2/config/Config.cpp
```diff
@@ -22,7 +22,3 @@ Copyright(c) 2020 Futurewei Cloud
 */
 
 #include "Config.h"
-
```
jfunston, line 25: Can remove this file now

src/k2/common/AutoGenFormattingUtils.h
```diff
@@ -0,0 +1,173 @@
...
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+    DAMAGES OR OTHER
+    LIABILITY,
+    WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+*/
+
+/* This entire file has been generated by common/gen_macros.py" */
```
jfunston, line 24: Should it be part of the build process to generate this?

src/k2/json/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+file(GLOB HEADERS "*.h")
+
+add_library(json INTERFACE)
```
jfunston, line 3: Do you still need to add json.hpp to the repo? It looks like you aren't using the macro the new version provides unless I missed it, so we can stick to the installed Ubuntu version.

jfunston (changes requested):

src/k2/common/AutoGenFormattingUtils.h
iafuture: fixed

src/k2/config/Config.cpp
iafuture: done

src/k2/common/FormattingUtils.h
iafuture: done

src/k2/cmd/tpcc/Log.h
iafuture: done

src/k2/json/CMakeLists.txt
iafuture: done

jfunston (approved):

PR #108, opened 2021-01-21
Fix TSO assert due to delayed heartbeat and adding more logging to de…

jerryhfeng:
…tect delay.

### DIFF
--- a/src/k2/tso/service/TSOController.cpp
+++ b/src/k2/tso/service/TSOController.cpp
@@ -219,6 +219,20 @@ void TSOService::TSOController::RegisterGetTSOWorkersURLs()
 
 void TSOService::TSOController::HeartBeat()
 {
+    // delayed heartbeat detection and logging
+    uint64_t nowCnt = now_nsec_count();
+
+    if ((nowCnt - _lastHeartBeat) > (uint64_t)( 2 * (nsec(_heartBeatTimerInterval()).count())))
+    {
+        if (_lastHeartBeat != 0)
+        {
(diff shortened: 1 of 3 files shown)


src/k2/tso/service/TSOController.cpp
```diff
@@ -219,6 +219,20 @@ void TSOService::TSOController::RegisterGetTSOWorkersURLs()
...
 {
+    // delayed heartbeat detection and logging
+    uint64_t nowCnt = now_nsec_count();
+
+    if ((nowCnt - _lastHeartBeat) > (uint64_t)( 2 * (nsec(_heartBeatTimerInterval()).count())))
+    {
+        if (_lastHeartBeat != 0)
+        {
+            K2LOG_I(log::tsoserver, "HeartBeat delayed more than two preset intervals, _lastHeartBeat:{}, preset duration in nano sec: {}", 
+                _lastHeartBeat,  nsec(_heartBeatTimerInterval()).count());
```
iafuture, line 230: you can just log _heartBeatTImerInterval(). I added conversion formatters for all timestamps as well so these should log nicely.

iafuture: .

iafuture (changes requested):

src/k2/tso/service/TSOController.cpp
iafuture: discussed - nanosec precision is needed so the change's good

iafuture (approved):

PR #123, opened 2021-01-31
Add a config to allow tso ignore reservedTimeThreshold. This threshol…

jerryhfeng:
…d value is extended with heartbeat, which can be delayed on single dev box

(where core is shared instead of dedicated). When it is not promptly updated/extended, TSO will choke. Adding the config to allow igore it.

### DIFF
--- a/src/k2/tso/service/TSOController.cpp
+++ b/src/k2/tso/service/TSOController.cpp
@@ -97,11 +97,15 @@ seastar::future<> TSOService::TSOController::InitializeInternal()
 
 void TSOService::TSOController::InitWorkerControlInfo()
 {
+    
     // initialize TSOWorkerControlInfo
+    _lastSentControlInfo.IgnoreThreshold =  _ignoreReservedTimeThreshold();
     _lastSentControlInfo.TBENanoSecStep =   seastar::smp::count - 1;            // same as number of worker cores
     _lastSentControlInfo.TsDelta =          _defaultTBWindowSize().count();     // uncertain window size of timestamp from the batch is also default _defaultTBWindowSize
     _lastSentControlInfo.BatchTTL =         _defaultTBWindowSize().count();     // batch's TTL is also _defaultTBWindowSize
 
+    _controlInfoToSend.IgnoreThreshold =    _ignoreReservedTimeThreshold();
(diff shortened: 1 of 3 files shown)


src/k2/tso/service/TSOController.cpp
```diff
@@ -294,9 +298,12 @@ seastar::future<> TSOService::TSOController::DoHeartBeat()
                 _myLease = std::get<0>(newLeaseAndThreshold);
                 _controlInfoToSend.ReservedTimeShreshold = std::get<1>(newLeaseAndThreshold);
 
-                uint64_t newCurTimeTSECount = TimeAuthorityNow();
-                K2ASSERT(log::tsoserver, _controlInfoToSend.ReservedTimeShreshold > newCurTimeTSECount && _myLease > newCurTimeTSECount,
-                    "new lease and ReservedTimeThreshold should be in the future.");
+                if (!_controlInfoToSend.IgnoreThreshold)
+                {
+                    uint64_t newCurTimeTSECount = TimeAuthorityNow();
+                    K2ASSERT(log::tsoserver, _controlInfoToSend.ReservedTimeShreshold > newCurTimeTSECount && _myLease > newCurTimeTSECount,
```
jerryhfeng, line 304: This assert is what currently still causing TSO fail to issue timestamp on dev box.

src/k2/tso/service/TSOController.cpp
```diff
@@ -372,7 +379,7 @@ seastar::future<> TSOService::TSOController::SendWorkersControlInfo()
         // worker can only issue TS under following condition
         if (!_stopRequested &&                                              // stop is not requested
             _prevReservedTimeShreshold < curTimeTSECount &&                 // _prevReservedTimeShreshold is in the past
-            _controlInfoToSend.ReservedTimeShreshold > curTimeTSECount )    // new ReservedTimeShreshold is in the future
+            (_controlInfoToSend.ReservedTimeShreshold > curTimeTSECount || _controlInfoToSend.IgnoreThreshold))    // new ReservedTimeShreshold is in the future OR it is ignored
         {
             readyToIssue = true;
```
jerryhfeng, line 384: This is another case that could cause TSO not issuing timestamp.

iafuture (approved):

PR #131, opened 2021-02-13
Tso - change to use rpc call interface(moving away from raw message interface), proper chosen endpoint based on client capability, reduce connection usage.

jerryhfeng:


src/k2/tso/client/tso_clientlib.cpp
```diff
@@ -87,42 +87,37 @@ seastar::future<> TSO_ClientLib::DiscoverServiceNodes(const k2::String& serverUR
...
+        .then([this](auto&& response) {
             if (_stopped) return seastar::make_ready_future<>();
 
-            if (!payload || payload->getSize() == 0)
+            auto& [status, r] = response;
+            if (!status.is2xxOK())
             {
-                K2LOG_E(log::tsoclient, "Remote end did not provide a data endpoint. Giving up");
-                return seastar::make_exception_future<>(std::runtime_error("no remote endpoint"));
+                K2LOG_E(log::tsoclient, "Error during get TSO node URLs, code:{}, error message:{}", status.code, status.message);
```
iafuture, line 99: you can just format status here instead of each part:  "nodeURLs, status={}", status)

src/k2/tso/client/tso_clientlib.cpp
```diff
@@ -87,42 +87,37 @@ seastar::future<> TSO_ClientLib::DiscoverServiceNodes(const k2::String& serverUR
...
-            K2ASSERT(log::tsoclient, !nodeURLs.empty(), "TSO server should have workers");
+            auto& nodeURLs = r.serviceNodeURLs;
+            if (nodeURLs.empty())
+            {
+                K2LOG_E(log::tsoclient, "Remote end did not provide node URLs. Giving up");
+                return seastar::make_exception_future<>(std::runtime_error("no remote endpoint"));
+            }
+            else
+            {
+                K2LOG_I(log::tsoclient, "received node URLs:{}", nodeURLs);
```
iafuture, line 112: i think this should be debug

src/k2/tso/client/tso_clientlib.cpp
```diff
@@ -87,42 +87,37 @@ seastar::future<> TSO_ClientLib::DiscoverServiceNodes(const k2::String& serverUR
...
-                    }
-                    else if (tempEndPoint.protocol == TCPRPCProtocol::proto)
-                    {
-                        // keep it to enPointToAdd, maybe replaced by RDMA endpoint later
-                        endPointToAdd = tempEndPoint;
-                    }
-                }
-                _curTSOServiceNodes.emplace_back(endPointToAdd);
+                _curTSOServiceNodes.push_back( Discovery::selectBestEndpoint(singleNodeURLs));
+                K2LOG_I(log::tsoclient, "Selected node endpoint:{}", _curTSOServiceNodes.back()->url);
```
iafuture, line 120: probably just debug here too

src/k2/tso/client/tso_clientlib.cpp
```diff
@@ -497,33 +492,35 @@ seastar::future<TimestampBatch> TSO_ClientLib::GetTimestampBatch(uint16_t batchS
...
                 {
                     K2LOG_I(log::tsoclient, "Stopping retry since we were stopped");
                     return seastar::make_exception_future<>(TSOClientLibShutdownException());
                 }
 
-                if (!replyPayload || replyPayload->getSize() == 0)
+                auto& [status, r] = response;
+                if (!status.is2xxOK())
+                {
+                    K2LOG_E(log::tsoclient, "Error during get timestampBatch, code:{}, error message:{}", status.code, status.message);
```
iafuture, line 513: same note on formatting status

src/k2/tso/client/tso_clientlib.cpp
```diff
@@ -497,33 +492,35 @@ seastar::future<TimestampBatch> TSO_ClientLib::GetTimestampBatch(uint16_t batchS
...
+                    // currently, we should only have 5xx retryable error
+                    K2ASSERT(log::tsoclient, status.is5xxRetryable(), "GetTimeStampBatch error should be 5xxRetryable.");
```
iafuture, line 515: you're throwing on the next line so this assertion I think won't be any different.  Maybe just remove the assertion?

src/k2/tso/service/TSOController.cpp
```diff
@@ -151,25 +151,26 @@ seastar::future<> TSOService::TSOController::GetAllWorkerURLs()
...
-        auto response = request.endpoint.newPayload();
-        response->write(_workersURLs);
-        return k2::RPC().sendReply(std::move(response), request);
-    });
+    K2LOG_D(log::tsoserver, "handleGetTSOServiceNodeURLs");
+    (void) request;
+
+    GetTSOServiceNodeURLsResponse response{.serviceNodeURLs = _workersURLs};
+    K2LOG_D(log::tsoserver, "returned TSO service nodes endpoints are: {}", _workersURLs);
+    return RPCResponse(Statuses::S200_OK("OK"), std::move(response));
```
iafuture, line 173: no need to specify a custom message unless you really want it in the status. just Statuses::S200_OK() would do too

src/k2/tso/service/TSOWorker.cpp
```diff
@@ -51,27 +54,29 @@ seastar::future<> TSOService::TSOWorker::gracefulStop()
...
-        return seastar::make_ready_future();
-    });
+        auto timestampBatchGot = GetTimestampFromTSO(request.batchSizeRequested);
+        GetTimeStampBatchResponse response{.timeStampBatch = timestampBatchGot};
+        K2LOG_D(log::tsoserver, "returned timeStampBatch: {}", timestampBatchGot);
+        return RPCResponse(Statuses::S200_OK("OK"), std::move(response));
+    }
+    catch(const TSONotReadyException& e)
+    {
+        K2LOG_E(log::tsoserver, "TSO Not Ready:{}", e.what());
```
iafuture, line 72: do we need these errors here?

iafuture (changes requested):

src/k2/tso/client/tso_clientlib.cpp
jerryhfeng: Good to know. Thx!

src/k2/tso/client/tso_clientlib.cpp
jerryhfeng: This is during initial setup, normally once per lift time. So far for such initialization, we have been using INFO level tracing, so I will keep it as it is now.

src/k2/tso/client/tso_clientlib.cpp
jerryhfeng: Same as it is in initlization.

src/k2/tso/client/tso_clientlib.cpp
jerryhfeng: The assertion here is just to make sure that future change (adding other error status) is handled. So I will keep it.

src/k2/tso/service/TSOWorker.cpp
jerryhfeng: Just want to have a clear distinction on unexpected error and known error case.

src/k2/tso/service/TSOController.cpp
jerryhfeng: Tried that, didn't work as () operator need a string param. Also tried directly Statuses::S200_OK, it doesn't work either as right value is needed here. Prefer not giving default empty string value to the () string param which will loose caller requirement to be explicit. So, leave as it is.

iafuture (approved):

PR #134, opened 2021-02-18
Doc k23 si states

iafuture:


iafuture (commented):
- should state how this doc fits with what we have. that is, the goal of the doc is to 
   - allow for persistence of txn state
   - implement improvement where we track txn states at each node
   - implement improvement where we perform async WI WAL persistence
   - rename states from Deleted->Finalized
 - it is a bit unclear what you have in mind about replay. Is state rebuilt on each replayed data record? Perhaps we need a short overview of how replay is proposed to work.
- maybe good time to rename deleted to finalized
- we should describe that DATA, TR, and MTR are three data record types which share the same WAL at each partition
- T1 doesn't need to be persisted
- forceAborted state for both in-memory and persisted
- the tables should talk about what we'd do in case of persistence failure (temporary and give-up after retries)
- There are state transitions missing from the table, e.g. if in T0.9 and receiving starting transaction(client is retrying)

- not sure if path(c) is correct. The decision to abort is only in-memory. if we crash and replay, we might actually proceed with the commit path. How do we ensure the result is consistent? (i.e. we always abort?)
- How does this fit with WI finalization
- what is in-memory state while T3/T5 is being done.
- should describe which actions can be done async vs which have to be done sync
- should describe which actions can be batched with others to avoid multiple persistence round trips (e.g. first WI is always batched with TR create and done async)
- Why introduce rename of ForceAborted to Obsolete too? I like ForceAborted better than Obsolete as it communicates that there was a forceful action making a transaction abort, but it is a staging point before the user-triggered Abort state. Also,  the state transitions you have are not correct here. Obsolete is a state out of which we come out either if client tells us to abort, or if we go outside RWE. We do not go into abort automatically based on heartbeat. Obsolete is also what we have after a push-based abort. We cannot go into ABORT after a push.
- I have a bit of a hard time tracking possible transitions since there are multiple tables. We should figure out how to present this in one table, and/or a diagram

PR #146, opened 2021-03-06
Heartbeat fix

jfunston:
The heartbeat.sh test failed 4 times in a row before the TxnManager heartbeat clock fix. Instrumentation showed  that the client was sending the heartbeat on time and it got a response from the server in under 100usec, so the problem was not overloading the network or future scheduling. After the TxnManager clock fix the test passed 4 times in a row.

### DIFF
--- a/src/k2/module/k23si/TxnManager.cpp
+++ b/src/k2/module/k23si/TxnManager.cpp
@@ -60,11 +60,15 @@ seastar::future<> TxnManager::start(const String& collectionName, dto::Timestamp
     _collectionName = collectionName;
     _hbDeadline = hbDeadline;
     updateRetentionTimestamp(rts);
+    // We need to call this now so that a recent time is used for a new
+    // transaction's heartbeat expiry if it comes in before the first heartbeat timer callback
+    CachedSteadyClock::now(true);
+
     _hbTimer.set_callback([this] {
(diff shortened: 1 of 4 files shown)


src/k2/module/k23si/TxnManager.cpp
```diff
@@ -64,7 +64,7 @@ seastar::future<> TxnManager::start(const String& collectionName, dto::Timestamp
         K2LOG_D(log::skvsvr, "txn manager check hb");
         _hbTask = _hbTask.then([this] {
             // refresh the clock
-            auto now = CachedSteadyClock::now(true);
+            auto now = Clock::now();
```
jerryhfeng, line 67: Why this fix the issue?

jerryhfeng (changes requested):

src/k2/module/k23si/TxnManager.cpp
jfunston: The problem was that the cachedclock is only updated on the hb timer callback. If a txn starts before the first HB callback is executed, its expiry date is very stale. The fix now is to refresh the cached clock on TxnManager::start

iafuture (approved):

jerryhfeng (approved):
Thanks!

PR #152, opened 2021-03-11
Idempotent write ops and DataRecord refactor

### LINKED ISSUE #116: Make rejectIfExists writes idempotent
If a write with the rejectIfExists flag is retried, the client may see a failure even if the first try placed a write intent. Make this operation idempotent by adding an op sequence id to write requests. This can be reused for future read-modify-write operations too.

jfunston:
Makes write ops idempotent, which fixes failures on rejectIfExist=true retries. Also refactors DataRecord to split committed records from write intents, and removes unnecessary data from committed records. Closes #116

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -342,7 +342,7 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
     for (; !_isScanDone(key_it, request, response.results.size());
                         _scanAdvance(key_it, request.reverseDirection, request.key.schemaName)) {
         auto& versions = key_it->second;
-        DataRecord* record = _getDataRecord(versions, request.mtr.timestamp);
+        DataRecord* record = _getDataRecordForRead(versions, request.mtr.timestamp);
         bool needPush = !record ? _needPush(versions, request.mtr.timestamp) : false;
 
         if (!record && !needPush) {
@@ -457,7 +457,7 @@ K23SIPartitionModule::handleRead(dto::K23SIReadRequest&& request, FastDeadline d
     }
 
     VersionSet& versions = IndexIt->second;
-    DataRecord* rec = _getDataRecord(versions, request.mtr.timestamp);
+    DataRecord* rec = _getDataRecordForRead(versions, request.mtr.timestamp);
     bool needPush = !rec ? _needPush(versions, request.mtr.timestamp) : false;
 
     // happy case: either committed, or txn is reading its own write, or there is no matching version
@@ -1166,7 +1166,7 @@ K23SIPartitionModule::_doPush(String collectionName, dto::Key key, dto::TxnId in
                         break;
                     }
                     case dto::TxnRecordState::Aborted: {
-                        versions.WI.reset(); // TODO-persistence: This shouldn't be done here but after successful persist, probably during txn finalization and/or GC for abandoned WIs
+                        _removeWI(IndexerIt);
                         break;
                     }
                     case dto::TxnRecordState::Committed: {
@@ -1304,7 +1304,7 @@ K23SIPartitionModule::handleTxnFinalize(dto::K23SITxnFinalizeRequest&& request) 
     // TODO-persistence: For now, remove aborted records right-away. With persistence we should do so after successfully
     // persisting
     if (versions.WI.has_value() && versions.WI->status == dto::WriteIntent::Status::Aborted) {
-        versions.WI.reset();
+        _removeWI(IndexerIt);
     }
 
     // send a partial update for updating the status of the record
@@ -1473,9 +1473,11 @@ bool K23SIPartitionModule::_needPush(const VersionSet& versions, const dto::Time
 // get the data record with the given key which is not newer than the given timestsamp, or if it
 // is an exact match for a write intent (for read your own writes, etc)
 dto::DataRecord*
-K23SIPartitionModule::_getDataRecord(VersionSet& versions, dto::Timestamp& timestamp) {
+K23SIPartitionModule::_getDataRecordForRead(VersionSet& versions, dto::Timestamp& timestamp) {
     if (versions.WI.has_value() && versions.WI->txnId.mtr.timestamp.compareCertain(timestamp) == 0) {
         return &(versions.WI->data);
+    } else if (versions.WI.has_value() && timestamp.compareCertain(versions.WI->txnId.mtr.timestamp) > 0) {
+        return nullptr;
     }
 
     auto viter = versions.committed.begin();
@@ -1490,6 +1492,16 @@ K23SIPartitionModule::_getDataRecord(VersionSet& versions, dto::Timestamp& times
     }
 
     return &(viter->data);
+}
+
+// Helper to remove a WI and delete the key from the indexer of there are no committed records
+void K23SIPartitionModule::_removeWI(IndexerIterator it) {
+    if (it->second.committed.size() == 0) {
+        _indexer.erase(it);
+        return;
+    }
+
+    it->second.WI.reset();
 }
 
 } // ns k2
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -343,7 +343,7 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
                         _scanAdvance(key_it, request.reverseDirection, request.key.schemaName)) {
         auto& versions = key_it->second;
         DataRecord* record = _getDataRecordForRead(versions, request.mtr.timestamp);
-        bool needPush = !record ? _needPush(versions, request.mtr.timestamp) : false;
+        bool needPush = !record ? _checkPushForRead(versions, request.mtr.timestamp) : false;
 
         if (!record && !needPush) {
             // happy case: we either had no versions, or all versions were newer than the requested timestamp
@@ -458,7 +458,7 @@ K23SIPartitionModule::handleRead(dto::K23SIReadRequest&& request, FastDeadline d
 
     VersionSet& versions = IndexIt->second;
     DataRecord* rec = _getDataRecordForRead(versions, request.mtr.timestamp);
-    bool needPush = !rec ? _needPush(versions, request.mtr.timestamp) : false;
+    bool needPush = !rec ? _checkPushForRead(versions, request.mtr.timestamp) : false;
 
     // happy case: either committed, or txn is reading its own write, or there is no matching version
     if (!needPush) {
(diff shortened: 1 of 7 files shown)


iafuture (changes requested):

src/k2/module/k23si/Module.cpp
```diff
@@ -1402,46 +1456,40 @@ K23SIPartitionModule::handleInspectAllKeys(dto::K23SIInspectAllKeysRequest&& req
...
 }
 
+// For a given challenger timestamp and key, check if a push is needed against a WI
+bool K23SIPartitionModule::_needPush(const VersionSet& versions, const dto::Timestamp& timestamp) {
+    if (!versions.WI.has_value()) {
+        return false;
+    }
+
+    // timestamps are unique, so if it is an exact match we know it is the same txn
+    if (versions.WI->txnId.mtr.timestamp.compareCertain(timestamp) == 0) {
```
jfunston, line 1466: need tsoid

test/k23si/3SITxnTest.cpp
```diff
@@ -1834,7 +1837,6 @@ seastar::future<> testScenario04() {
                         return doInspectRecords(k2, collname)
                         .then([&](auto&& response)  {
                             auto& [status, val] = response;
-                            K2EXPECT(log::k23si, status, dto::K23SIStatus::KeyNotFound);
```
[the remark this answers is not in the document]
jfunston: Added back in the ability to remove a key from the indexer and added this check back in.

test/k23si/3SITxnTest.cpp
```diff
@@ -2485,10 +2479,6 @@ seastar::future<> testScenario07() {
                         auto [status2, val2] = resp2.get0();
                         K2EXPECT(log::k23si, status1, dto::K23SIStatus::OK);
                         K2EXPECT(log::k23si, status2, dto::K23SIStatus::OK);
-                        K2EXPECT(log::k23si, val1.records[0].txnId.mtr.txnid, mtr.txnid);
```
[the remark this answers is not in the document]
jfunston: Added back in some of the checks that I could

test/k23si/3SITxnTest.cpp
```diff
@@ -2377,14 +2378,7 @@ seastar::future<> testScenario06() {
                 K2EXPECT(log::k23si, status2, dto::K23SIStatus::Created);
             })
             .then([&] {
-                return doFinalize(k8, k8, mtr, collname, true, ErrorCaseOpt::NoInjection)
```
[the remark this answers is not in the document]
jfunston: Changed this test to use an invalid MTR instead of an invalid TRH key.

src/k2/module/k23si/Module.cpp
jfunston: Checked the code and compareCertain does use the tsoID

src/k2/module/k23si/Module.cpp
```diff
@@ -342,17 +342,18 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
     for (; !_isScanDone(key_it, request, response.results.size());
                         _scanAdvance(key_it, request.reverseDirection, request.key.schemaName)) {
         auto& versions = key_it->second;
-        auto viter = _getVersion(versions, request.mtr.timestamp);
+        DataRecord* record = _getDataRecord(versions, request.mtr.timestamp);
+        bool needPush = !record ? _needPush(versions, request.mtr.timestamp) : false;
```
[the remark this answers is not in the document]
jfunston: getDataRecord was only used for read path so renamed it to make it more clear. Fixed this bug by making it return a nullptr if there is a WI from a different txn and earlier timestamp.

src/k2/module/k23si/Module.cpp
```diff
@@ -887,28 +895,38 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, FastDeadline
...
-                    if (retryChallenger) {
-                        K2LOG_D(log::skvsvr, "Partition: {}, write push retry for key {}", _partition, request.key);
-                        return handleWrite(std::move(request), deadline);
-                    }
-                    // challenger must fail
-                    K2LOG_D(log::skvsvr, "Partition: {}, write push challenger lost for key {}", _partition, request.key);
-                    return RPCResponse(dto::K23SIStatus::AbortConflict("incumbent txn won in write push"), dto::K23SIWriteResponse{});
-                });
-        }
+    if (versions.WI.has_value() && versions.WI->txnId.mtr != request.mtr) {
```
iafuture, line 898: I think this should only be done for IPR WI. If the WI is aborted, we should probably remove it from memory

src/k2/module/k23si/Module.cpp
```diff
@@ -887,28 +895,38 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, FastDeadline
...
+                // challenger must fail
+                K2LOG_D(log::skvsvr, "Partition: {}, write push challenger lost for key {}", _partition, request.key);
+                return RPCResponse(dto::K23SIStatus::AbortConflict("incumbent txn won in write push"), dto::K23SIWriteResponse{});
+            });
     }
 
-    if (request.rejectIfExists && versions.size() > 0 && !versions[0].isTombstone) {
+    // Handle idempotency here. If request ids match, then this was a retry message from the client
+    // and we should return OK
+    if (versions.WI.has_value() && request.mtr == versions.WI->txnId.mtr &&
```
iafuture, line 916: same as above - i think this only applies for IPR WI

src/k2/module/k23si/Module.cpp
```diff
@@ -499,15 +506,16 @@ Status K23SIPartitionModule::_validateStaleWrite(const RequestT& request, Versio
     // if a txn committed a value at time T5, then we must also assume they did a read at time T5
     // NB(3) if we encounter a WI, we check the second oldest version to see if there is a need to push.
     // If the second oldest is newer than we are, then we won't commit even if we win a PUSH against the WI.
```
iafuture, line 508: Should we consider the op ID now for stale write? I.e. lower sequence# should be considered stale against a higher one

src/k2/module/k23si/Module.cpp
```diff
@@ -1219,42 +1233,78 @@ K23SIPartitionModule::handleTxnFinalize(dto::K23SITxnFinalizeRequest&& request)
...
+            return RPCResponse(dto::K23SIStatus::OK("finalize key missing in abort"), dto::K23SITxnFinalizeResponse());
+        }
+        // Normal abort case
+        else if (found && isWI) {
+            K2LOG_D(log::skvsvr, "Partition: {}, aborting {}, in txn {}", _partition, request.key, txnId);
+            versions.WI->status = dto::WriteIntent::Status::Aborted;
+        }
+        // Error case, trying to abort but it was already committed
+        else {
+            K2LOG_D(log::skvsvr, "Partition: {}, cannot abort committed record", _partition);
```
iafuture, line 1271: We should extend the message in the response to explain why we can't finalize

src/k2/module/k23si/Module.cpp
```diff
@@ -1219,42 +1233,78 @@ K23SIPartitionModule::handleTxnFinalize(dto::K23SITxnFinalizeRequest&& request)
...
-            K2LOG_D(log::skvsvr, "Partition: {}, cannot commit aborted record: {}", _partition, *rec);
-            return RPCResponse(dto::K23SIStatus::OperationNotAllowed("cannot finalize txn"), dto::K23SITxnFinalizeResponse());
-        default:
-            // the action did not match the state
-            K2LOG_D(log::skvsvr,
-                "Partition: {}, failing finalize due to action mismatch {}, in txn {}, have status={}, asked={}",
-                _partition, request.key, txnId, rec->status, request.action);
+    }
+
+    if (request.action == dto::EndAction::Abort) {
```
iafuture, line 1258: I think this block would read better if you used a switch instead of another if-elseif layer

src/k2/module/k23si/Module.cpp
```diff
@@ -1219,42 +1233,78 @@ K23SIPartitionModule::handleTxnFinalize(dto::K23SITxnFinalizeRequest&& request)
...
+            return RPCResponse(dto::K23SIStatus::OperationNotAllowed("cannot commit missing key"),
+                               dto::K23SITxnFinalizeResponse());
+        }
+    }
+    // Error case, something wrong with the request
+    else {
+        K2LOG_D(log::skvsvr,
+            "Partition: {}, failing finalize due to action mismatch {}, in txn {}, asked={}",
+            _partition, request.key, txnId, request.action);
+        return RPCResponse(dto::K23SIStatus::OperationNotAllowed("cannot finalize txn"), dto::K23SITxnFinalizeResponse());
```
iafuture, line 1298: a bit better message here - why can't we finalize?

iafuture (changes requested):

src/k2/module/k23si/Module.cpp
```diff
@@ -1402,46 +1456,52 @@ K23SIPartitionModule::handleInspectAllKeys(dto::K23SIInspectAllKeysRequest&& req
...
 }
 
-// get the data record with the given key which is not newer than the given timestsamp
+// get the data record with the given key which is not newer than the given timestsamp, or if it
+// is an exact match for a write intent (for read your own writes, etc)
 dto::DataRecord*
-K23SIPartitionModule::_getDataRecord(const dto::Key& key, const dto::Timestamp& timestamp) {
-    auto versions = _indexer.find(key);
-    if (versions == _indexer.end()) {
+K23SIPartitionModule::_getDataRecordForRead(VersionSet& versions, dto::Timestamp& timestamp) {
```
jfunston, line 1476: read with timestamp lower than any committed record

src/k2/module/k23si/Module.cpp
```diff
@@ -1237,76 +1232,81 @@ K23SIPartitionModule::handleTxnFinalize(dto::K23SITxnFinalizeRequest&& request)
...
     if (versions.WI.has_value() && versions.WI->txnId == txnId) {
         found = true;
         isWI = true;
-        isAborted = versions.WI->status == dto::WriteIntent::Status::Aborted;
     }
     // Check for a matching committed record if needed
     if (!found) {
         for (const CommittedRecord& record : versions.committed) {
             // timestamps are unique, so they can identify a txn match
             if (record.timestamp.compareCertain(txnId.mtr.timestamp) == 0) {
```
iafuture, line 1245: I think it'd be slightly more efficient(fewer branches) to compare once and break, something like
``` c++
int comp = record.timetamp.compareCertain(txnId.mtr.timestamp);
found = !comp;
if (comp <= 0) break;
```

iafuture (changes requested):

iafuture (approved):

PR #160, opened 2021-04-05
Drop collection

### LINKED ISSUE #138: SKV support for drop collection

jfunston:
Closes #138


docs/RFC/drop_collection.md
```diff
@@ -0,0 +1,47 @@
...
+- CPO moves the collection metadata to a deleted list. The collection will no longer be returned as part of get collection requests. Its schemas are deleted.
+- CPO issues unload partition request to all collection partitions
+- A server node receiving an unload partition request will immediately drop the partition, without waiting for ongoing operations and with no persistence needed.
+- After receiving responses from all partitions, CPO responds to the client.
+- From this point on, a new collection with the same name may be recreated.
+- Asynchronously, the CPO cleans the WAL persistence data (not implemented yet)
+
+### Notes
+
+The CPO is the authoratative truth of collection metadata such as the partition map, version, and the
```
iafuture, line 15: you should let MD format the paragraphs instead of manual \n's

src/k2/assignmentManager/AssignmentManager.cpp
```diff
@@ -63,21 +66,71 @@ AssignmentManager::handleAssign(dto::AssignmentCreateRequest&& request) {
...
+    if (!_pmodule || _collectionName != request.collectionName) {
+        return RPCResponse(Statuses::S403_Forbidden("offload not allowed"), dto::AssignmentOffloadResponse{});
+    }
+
     K2LOG_I(log::amgr, "Received request to offload assignment for {}", request.collectionName);
-    return RPCResponse(Statuses::S501_Not_Implemented("offload has not been implemented"), dto::AssignmentOffloadResponse());
+
+    return _pmodule->gracefulStop()
+    .then([this] () {
+        _pmodule = nullptr;
```
iafuture, line 131: can you add Info logs before/after this call?

src/k2/cpo/service/CPOService.cpp
```diff
@@ -383,6 +411,44 @@ void CPOService::_assignCollection(dto::Collection& collection) {
         }));
 }
 
+seastar::future<> CPOService::_offloadCollection(dto::Collection& collection) {
```
iafuture, line 414: you don't have any retries here(which is fine) but we should make sure we can retry from the outside if things time out/fail.

I think to achieve that we need:
- 404_not_found response from node if the collection has already been offloaded there
- handling of 404 here as a non-error condition (along with the 200 ok)
- we should return any other error status out and only clean up the cpo state if everything succeeded
- we should be able to mark the cpo state as "deleted" when we receive a drop request. That way we can still have the metadata around but we know not to return it to clients. The metadata should be here until all partitions ACK the drop. If any fail, we leave it untouched and let the client retry if they want

iafuture (changes requested):

iafuture (approved):

iafuture (approved):

src/k2/cpo/service/CPOService.cpp
```diff
@@ -215,20 +222,59 @@ CPOService::handleGet(dto::CollectionGetRequest&& request) {
...
 
+seastar::future<std::tuple<Status, dto::CollectionDropResponse>>
+CPOService::handleCollectionDrop(dto::CollectionDropRequest&& request) {
+    K2LOG_I(log::cposvr, "Received collection drop request for {}", request.name);
+    auto [status, collection] = _getCollection(request.name);
+
+    if (!status.is2xxOK()) {
+        return RPCResponse(std::move(status), dto::CollectionDropResponse());
+    }
+
```
jerryhfeng, line 245: check deleted flag here for re-try drop request?

src/k2/cmd/nodepool/nodepool_main.cpp
```diff
@@ -47,7 +46,6 @@ int main(int argc, char** argv) {
     app.addApplet<k2::CollectionMetadataCache>();
     app.addApplet<k2::NodePoolMonitor>();
     app.addApplet<k2::AssignmentManager>();
-    app.addApplet<k2::PartitionManager>();
```
jerryhfeng, line 50: Why removing this? Leave us no possibility to host multiple partition (of same collection) in one node in the future?

src/k2/partitionManager/PartitionManager.h
```diff
@@ -1,55 +0,0 @@
...
-#include <k2/module/k23si/Module.h>
-#include <seastar/core/distributed.hh>  // for dist stuff
-#include <seastar/core/future.hh>       // for future stuff
-
-namespace k2 {
-namespace log {
-inline thread_local k2::logging::Logger partmgr("k2::partition_manager");
-}
-
-class PartitionManager {
```
jerryhfeng, line 38: Same concern about possible support of multiple partitions in the future on one node.

jerryhfeng (approved):

PR #162, opened 2021-04-07
Add a couple record erase tests

jfunston:


test/k23si/QueryTest.cpp
```diff
@@ -624,6 +625,61 @@ seastar::future<> runScenario06() {
...
+    .then([this] () {
+        return writeTxn.end(true);
+    })
+    .then([] (auto&& response) {
+        K2EXPECT(log::k23si, response.status, k2::dto::K23SIStatus::OK);
+        return seastar::make_ready_future<>();
+    })
+    .then([this] () {
+        return doQuery("scenario07", "scenario07", -1, false, 0, 1).discard_result();
+    });
```
iafuture, line 680: Is the point here that you get a 200 response on the query with no records returned? I think you should assert that after doQuery

iafuture (changes requested):

test/k23si/QueryTest.cpp
jfunston: doQuery is a test method that does that already, and it asserts that the number of records returned is as expected, 0 in this case.

iafuture (approved):

Issue #74, opened 2020-10-27
Add skip and copy methods to Payload


jfunston:
This would reduce data copies in the code that needs this functionality (e.g. in partial update code in Module.cpp)

jerryhfeng:
Curious if the compiler could optimize it away. But even so, it still worth to make the code change for better readability/intention.

iafuture:
it'd be nice to have an api on Payload to help copy raw data from one payload to another (e.g. during projection)
Proposed api:
``` c++
template <typenameT>
bool copy<T>(Payload& dst);
```

PR #164, opened 2021-04-12
Add copy & skip template method in payload (#74)

zsStrike:


jfunston:
Hi, thanks for the PR, we will discuss more in this week's meeting but I wanted to give some written feedback as well so that the meeting can be as efficient as possible.

First, I think that the key point of both the copy and skip methods is knowing how many bytes are used to serialize into a payload for each type. So if we had a method on the payload like:
`
template <typename T>
`
`
static size_t getSerializedSizeOf();
`
Then you can use this method to implement both the skip and copy methods of the payload and each of those will only require a few lines of code. Only the getSerializedSizeOf() method will need specializations for each type. When writing getSerializedSizeOf you should implement it by using template specialization instead of function overload like you have now. A simple example of template specialization in our code is the TToFieldType function in FieldTypes.h and FieldTypes.cpp. Doing it this way makes the code cleaner and means that you don't have to initialize temporary values.

You also need to support all of the types that read and write on the Payload supports. For this you need to support structs with the K2_PAYLOAD_FIELDS and K2_PAYLOAD_COPYABLE macros defined, and you will need to modify K2_PAYLOAD_FIELDS macro so that it adds a method to the struct that gets the size of the member variables and can work recursively for nested structs with the macro defined. Your test cases should include some complicated nested structs like this along with container types for both skip and copy.

zsStrike:
Thanks for your detailed advice, it is quiet helpful to me. I will implement the new copy & skip method later.

iafuture (approved):
Looks great to me!

PR #165, opened 2021-04-14
Add getSKVKeyRecord method

jfunston:


src/k2/dto/SKVRecord.cpp
```diff
@@ -250,6 +251,45 @@ SKVRecord SKVRecord::cloneToOtherSchema(const String& collection, std::shared_pt
...
+
+    size_t num_keys = schema->partitionKeyFields.size() + schema->rangeKeyFields.size();
+    SKVRecord::Storage key_storage = storage.share();
+
+    if (key_storage.excludedFields.size() == 0) {
+        key_storage.excludedFields.resize(schema->fields.size());
+        for (size_t i = 0; i < num_keys; ++i) {
+            key_storage.excludedFields[i] = false;
+        }
+    }
```
iafuture, line 276: A little confused what this is trying to achieve. I think the eventual goal is to have a bitmap which marks all non-key fields as `excluded`, and all non-previously-excluded key fields as `included`. I'd suggest a bit more compact form to achieve that:

``` c++
// mark everything as excluded
key_storage.excludedFields.clear();
key_storage.excludedFields.resize(schema->fields.size(), true);
for (size_t i = 0; i < num_keys; ++i) {
    if (!storage.excludedFields[i]) {
        K2_DTO_CAST_APPLY_FIELD_VALUE(skipFieldDataHelper, schema->fields[i], key_storage.fieldData);
    }
}
```

iafuture (approved):

PR #186, opened 2021-06-17
Resolved Issue 170

### LINKED ISSUE #170: Disallow NaN for storage of float, double, and decimals
Detect and forbid NaN values in SKVRecord serializeField and in filter expression evaluation. Any NaN comparison results in false and we want to avoid these unintuitive situations and have the option in the future to have these types as key fields.

ammuv:
Disallowed NaN for storage of float and double. 

Closes #170

### DIFF
--- a/src/k2/dto/SKVRecord.h
+++ b/src/k2/dto/SKVRecord.h
@@ -65,20 +65,40 @@ struct KeyNotAvailableError : public std::exception {
     virtual const char* what() const noexcept override { return what_str.c_str(); }
 };
 
+template <typename T>
+bool isNan(const T& field){
+    if constexpr (std::is_same_v<T, float> || std::is_same_v<T, double> )  { // handle NaN float and double
+        if (std::isnan(field)) {
+            return true;
+        }
+    }
+
+    if constexpr (std::is_same_v<T, std::decimal::decimal64>)  { // handle NaN decimal
+        if (std::isnan(std::decimal::decimal64_to_float(field))) {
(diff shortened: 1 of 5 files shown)


src/k2/dto/SKVRecord.h
```diff
@@ -59,16 +65,42 @@ struct KeyNotAvailableError : public std::exception {
     virtual const char* what() const noexcept override { return what_str.c_str(); }
 };
 
+template <typename T>
+bool isNan(const T& field){
```
jfunston, line 69: I think this would fit better in FieldTypes.h. Can you move it there and add a comment that we use it to prevent NaNs in fields?

src/k2/dto/SKVRecord.h
```diff
@@ -59,16 +65,42 @@ struct KeyNotAvailableError : public std::exception {
...
+        if(isNan<T>(field)){
+            throw NaNError("NaN type in serialization");
+        }
+
         FieldType ft = TToFieldType<T>();
         if (fieldCursor >= schema->fields.size() || ft != schema->fields[fieldCursor].type) {
             throw TypeMismatchException("Schema not followed in record serialization");
         }
-
+
```
jfunston, line 103: Some extra whitespace here. I think Ivan has some Visual Studio Code settings to fix this automatically, you can ask him about it.

jfunston (changes requested):

jfunston (approved):

PR #197, opened 2021-07-14
retry failed transcations in TPCC benchmark tests

johnfangAFW:
1) changed origin run() as attempt()
2) added retry strategy to try for a number of attempts.
3) tpcc integration tests succeeded

./test_k23si_tpcc.sh

### DIFF
--- a/src/k2/cmd/tpcc/transactions.h
+++ b/src/k2/cmd/tpcc/transactions.h
@@ -37,6 +37,43 @@ Copyright(c) 2020 Futurewei Cloud
 using namespace seastar;
 using namespace k2;
 
+// a simple retry strategy, stop after a number of retries
+class FixedRetryStrategy {
+public:
+    FixedRetryStrategy(int retries) : _retries(retries), _try(0), _success(false) {
+    }
+
+    ~FixedRetryStrategy() = default;
+
+template<typename Func>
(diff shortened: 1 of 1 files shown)


src/k2/cmd/tpcc/transactions.h
```diff
@@ -50,8 +87,18 @@ class AtomicVerify;
 
 class TPCCTxn {
 public:
-    virtual future<bool> run() = 0;
+    virtual future<bool> attempt() = 0;
     virtual ~TPCCTxn() = default;
+
+    future<bool> run() {
+        // retry 10 times for a failed transaction
+        auto retryStrategy = make_shared<FixedRetryStrategy>(10);
```
jfunston, line 95: You can make this a unique_ptr and capture it by move in a "finally" block to safely dispose of it.

jfunston (changes requested):

src/k2/cmd/tpcc/transactions.h
iafuture: I was going to suggest to use a do_with block here instead of the pointer create and capture trick

johnfangAFW: changed to use do_with block as suggested.

jfunston (approved):

PR #200, opened 2021-07-16
add transaction weights to TPCC client

johnfangAFW:
Introduced an option --txn_weights for tpcc client, for example,

./build/src/k2/cmd/tpcc/tpcc_client -c1 --tcp_remotes ${EPS} --cpo ${CPO} --tso_endpoint ${TSO} --num_warehouses ${NUMWH} --districts_per_warehouse ${NUMDIST} --prometheus_port 63101 ${COMMON_ARGS} --memory=512M --partition_request_timeout=1s  --num_concurrent_txns=1 --do_verification false --delivery_txn_batch_size=10 --txn_weights={43,4,4,45,4}

Updated integration test script and doc. 

TPCC integration tests passed.

### DIFF
--- a/src/k2/cmd/tpcc/tpcc_client.cpp
+++ b/src/k2/cmd/tpcc/tpcc_client.cpp
@@ -109,6 +109,8 @@ public:  // application lifespan
     seastar::future<> start() {
         _stopped = false;
 
+        _weights = _parse_weights(_txn_weights());
+
         setupSchemaPointers();
 
         registerMetrics();
@@ -149,6 +151,23 @@ public:  // application lifespan
     }
 
 private:
+    std::vector<uint32_t> _parse_weights(k2::String txn_weights) {
+        K2LOG_D(log::tpcc, "Parsing transaction weights {}", txn_weights);
+        std::stringstream ss(txn_weights);
+        std::string weight_str;
+        std::vector<uint32_t> weights;
+
+        uint32_t sum = 0;
+        while (getline (ss, weight_str, ',')) {
+            uint32_t weight = std::stoi(weight_str);
+            sum += weight;
+            weights.push_back(sum);
+        }
+        K2ASSERT(log::tpcc, weights.size() == 5, "number of transaction types should be 5, but the actual is {}", weights.size());
+        K2ASSERT(log::tpcc, sum == 100, "The sum of transaction weights should be 100, but the actual value is {}", sum);
+        return weights;
+    }
+
     seastar::future<> _schema_load() {
         std::vector<seastar::future<>> schema_futures;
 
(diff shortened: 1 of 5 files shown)


src/k2/cmd/tpcc/tpcc_client.cpp
```diff
@@ -149,6 +151,23 @@ class Client {
     }
 
 private:
+    std::vector<uint32_t> _parse_weights(k2::String txn_weights) {
```
iafuture, line 154: I don't think you need a custom parser for this. You can just define the option in config as:
```c++
("txn_weights", bpo::value<std::vector<int>>()->multitoken()->default_value(std::vector<int>()), "A space-delimited list of exactly 5 elements denoting the percentage for each txn type:" <list txn types here>)
```

and then pull it out with a configVar:
```c++
    ConfigVar<std::vector<int>> _txn_weights{"txn_weights"};
```

src/k2/cmd/tpcc/tpcc_client.cpp
```diff
@@ -246,15 +265,15 @@ class Client {
                 uint32_t w_id = (seastar::this_shard_id() % _max_warehouses()) + 1;
                 uint32_t d_id = (seastar::this_shard_id() % _districts_per_warehouse()) + 1;
                 TPCCTxn* curTxn;
-                if (txn_type <= 43) {
+                if (txn_type <= _weights[0]) {
```
iafuture, line 268: I don't think this is equivalent to the previous code. If your first txn is of largest weight (say 51), then you will never run any other transactions.

You should add up the weights to 100 and then pick a txn type based on where the random number falls (as the previous code did)

src/k2/cmd/tpcc/tpcc_client.cpp
```diff
@@ -350,6 +369,7 @@ class Client {
     ConfigVar<int> _num_concurrent_txns{"num_concurrent_txns"};
     ConfigVar<uint16_t> _delivery_txn_batch_size{"delivery_txn_batch_size"};
     ConfigVar<uint16_t> _districts_per_warehouse{"districts_per_warehouse"};
+    ConfigVar<k2::String> _txn_weights{"txn_weights"};
```
iafuture, line 372: related to the parser above - just grab it as a vector<int>

iafuture (changes requested):

src/k2/cmd/tpcc/tpcc_client.cpp
johnfangAFW: nice, didn't know this feature. will do.

src/k2/cmd/tpcc/tpcc_client.cpp
johnfangAFW: the logic should be the same since the _weights actually holds the partial sum for each iteration in the parsing logic instead of the weight itself.

src/k2/cmd/tpcc/tpcc_client.cpp
johnfangAFW: will do

johnfangAFW: done

src/k2/cmd/tpcc/tpcc_client.cpp
johnfangAFW: done

jfunston (approved):

iafuture (approved):

PR #201, opened 2021-07-20
add customer name secondary index to TPCC

johnfangAFW:
Implemented method getCIdByLastNameViaIndex() to replace getCIdByLastName().

TPCC integration tests succeeded.

### DIFF
--- a/src/k2/cmd/tpcc/transactions.h
+++ b/src/k2/cmd/tpcc/transactions.h
@@ -146,7 +146,7 @@ private:
     future<bool> runWithTxn() {
         future<> warehouse_update = warehouseUpdate();
         future<> district_update = districtUpdate();
-        future<> CId_get = getCIdByLastName();
+        future<> CId_get = getCIdByLastNameViaIndex();
 
         return when_all_succeed(std::move(warehouse_update), std::move(district_update), std::move(CId_get)).discard_result()
         .then([this] () {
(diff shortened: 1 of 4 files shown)


src/k2/cmd/tpcc/transactions.h
```diff
@@ -241,8 +241,72 @@ class PaymentT : public TPCCTxn
...
+                    return make_ready_future();
+                });
+            });
+        });
+    }
+
     // get customer ID by last name
-    // 60% select customer by last name; 40%, the _c_id has already been set randomly at the consturctor
+    // 60% select customer by last name; 40%, the _c_id has already been set randomly at the constructor
     future<> getCIdByLastName() {
```
jfunston, line 310: You can remove this function now

src/k2/cmd/tpcc/transactions.h
```diff
@@ -241,8 +241,72 @@ class PaymentT : public TPCCTxn
         return writeRow<History>(history, _txn).discard_result();
     }
 
+    future<> getCIdByLastNameViaIndex() {
+        uint32_t cid_type = _random.UniformRandom(1, 100);
+        if (cid_type > 60) {
```
jfunston, line 246: It should be 60% chance to select by last name, so this should be <= 60. This bug existed in the previous code but we might as well fix it now.

src/k2/cmd/tpcc/transactions.h
```diff
@@ -241,8 +241,72 @@ class PaymentT : public TPCCTxn
         return writeRow<History>(history, _txn).discard_result();
     }
 
+    future<> getCIdByLastNameViaIndex() {
```
jfunston, line 244: The order status transaction also selects customer by last name and can use the secondary index.

jfunston (changes requested):

src/k2/cmd/tpcc/transactions.h
johnfangAFW: removed, thanks

src/k2/cmd/tpcc/transactions.h
johnfangAFW: fixed and changed it to <=60

src/k2/cmd/tpcc/transactions.h
johnfangAFW: replaced getCustomerId() with getCustomerIdViaIndex() for the order status transaction.

jfunston (approved):

PR #210, opened 2021-08-12
issue 120

raduvine:


src/k2/dto/SKVRecord.h
```diff
@@ -75,24 +79,25 @@ class SKVRecord {
...
         for (size_t i = 0; i < schema->rangeKeyFields.size(); ++i) {
-            if (schema->rangeKeyFields[i] == fieldCursor) {
+            if (schema->rangeKeyFields[i] == serializerFieldCursor) {
                 rangeKeys[i] = FieldToKeyString<T>(field);
             }
         }
 
         storage.fieldData.write(field);
-        ++fieldCursor;
+        ++serializerFieldCursor;
```
jfunston, line 100: In this case we have advanced the underlying payload cursor, so deserialzeFieldCursor should be incremented as well. Otherwise the user might expect to be able to deserialize the value just serialized, but that isn't possible without calling seekField first. In general, the deserializeFieldCursor should always match the last field serialized regardless of the payload cursor and the serializeFieldCursor should match the payload cursor. So calling serializeNext may require calling seekField and will necessarily require incrementing the deserializeFieldCursor.

jfunston: We should also have a unit tests for serializing and deserializing the same SKVRecord which will test the logic in the above comment.

src/k2/dto/SKVRecord.h
```diff
@@ -182,7 +187,7 @@ class SKVRecord {
     // no-arg version to satisfy the template expansion above in the terminal case
     void readMany() {}
 
-    uint32_t getFieldCursor() const;
+    uint32_t getFieldCursor(CursorType cursorType) const;
```
jfunston, line 190: We should have a default argument for this parameter that is the deserializerFieldCursor so that other code, like chogori-sql, is still compatible

src/k2/dto/SKVRecord.h
```diff
@@ -252,7 +257,8 @@ class SKVRecord {
 
     std::vector<String> partitionKeys;
     std::vector<String> rangeKeys;
-    uint32_t fieldCursor = 0;
+    uint32_t serializerFieldCursor = 0;    // field possition during serialization
+    uint32_t deserializerFieldCursor = 0;  // field possition during deserialization
```
jfunston, line 261: There is a constructor for the SKVRecord that takes a Storage argument. In that constructor deserializeFieldCursor should be initialized to the end, since if the SKVRecord is constructed from an existing Storage then all the fields are already serialized.

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -279,10 +279,16 @@ class K2TxnHandle {
 
         std::unique_ptr<dto::K23SIWriteRequest> request;
         if constexpr (std::is_same<T, dto::SKVRecord>()) {
+            // check if all fields have been serialized
+            if (record.getFieldCursor(k2::dto::SKVRecord::SERIALIZER) != record.schema->fields.size())
```
jfunston, line 283: We should have an integration test in the SKVClient tests for this case.

jfunston (changes requested):

PR #213, opened 2021-08-18
One hot transaction

ammuv:
- Added one hot transaction feature to K23SI client and server.
- Modified YCSB tests to use one hot transactions.

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -1108,11 +1108,40 @@ K23SIPartitionModule::_processWrite(dto::K23SIWriteRequest&& request, FastDeadli
         }
     }
 
+    // build write_ranges for Transaction end request
+    std::unordered_map<String, std::unordered_set<dto::KeyRangeVersion>> write_ranges; // will be filled in later if WI is created successfully
 
-    // all checks passed - we're ready to place this WI as the latest version
-    auto status = _createWI(std::move(request), vset);
-    K2LOG_D(log::skvsvr, "WI creation with status {}", status);
-    return RPCResponse(std::move(status), dto::K23SIWriteResponse{});
+    //endRequest created for calling EndTransaction for one hot transactions.
+    dto::K23SITxnEndRequest endRequest{request.pvid,
+        request.trhCollection,
+        request.trh,
+        request.mtr,
+        dto::EndAction::Commit,
+        std::move(write_ranges)
+    };
+
+    return seastar::do_with((dto::K23SITxnEndRequest)std::move(endRequest), (bool)request.isonehot,
+                            [this, &request] (dto::K23SITxnEndRequest& endRequest, bool& isonehot) {
+        // all checks passed - we're ready to place this WI as the latest version
+        auto& vset = _indexer[request.key];
+        auto status = _createWI(std::move(request), vset);
+        K2LOG_D(log::skvsvr, "WI creation with status {}", status);
+
+        // not one hot transaction or WI creation failed
+        if(!isonehot || !status.is2xxOK()) {
+            return RPCResponse(std::move(status), dto::K23SIWriteResponse{});
+        }
+
+        endRequest.writeRanges[request.collectionName].insert(_partition().keyRangeV); // add keyRangeV of current partition (TRH)
+
+        // End one hot transaction
+        return handleTxnEnd(std::move(endRequest))
+                .then( [this] (auto&& response) {
+                    auto& [endstatus, k2response] = response;
+                    return RPCResponse(std::move(endstatus), dto::K23SIWriteResponse{}); // Return status after ending one hot transaction
+                });
+
+    });
 }
 
 Status
--- a/src/k2/module/k23si/client/k23si_client.cpp
+++ b/src/k2/module/k23si/client/k23si_client.cpp
@@ -120,6 +120,7 @@ seastar::future<ReadResult<dto::SKVRecord>> K2TxnHandle::read(dto::Key key, Stri
 
     _client->read_ops++;
     _ongoing_ops++;
+    _total_ops++;
 
     return _cpo_client->partitionRequest
         <dto::K23SIReadRequest, dto::K23SIReadResponse, dto::Verbs::K23SI_READ>
@@ -156,7 +157,7 @@ seastar::future<ReadResult<dto::SKVRecord>> K2TxnHandle::read(dto::Key key, Stri
 
 
 std::unique_ptr<dto::K23SIWriteRequest> K2TxnHandle::_makeWriteRequest(dto::SKVRecord& record, bool erase,
-                                                                    dto::ExistencePrecondition precondition) {
+                                                                    dto::ExistencePrecondition precondition, bool isonehot=false) {
     for (const String& key : record.partitionKeys) {
         if (key == "") {
             throw K23SIClientException("Partition key field not set for write request");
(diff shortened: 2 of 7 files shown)


iafuture (commented):
Initially i thought it was a typo, but what is "one hot txn"? I think it should be "one-shot" or "one-hop"

ammuv:
> Initially i thought it was a typo, but what is "one hot txn"? I think it should be "one-shot" or "one-hop"

Oh...I thought it was one-hot the whole time. Like they are "hot" transactions or transactions that are quicker because commits need not be sent by client. But now that you say, "shot" makes more sense!

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -370,23 +407,40 @@ class K2TxnHandle {
...
+                    }
+                    //note we need not end heartbeat timer because we never started it for one hot transactions.
+                    auto time_spent = Clock::now() - _start_time;
+                    if (time_spent < 5us) {
+                        sleep = 5us - time_spent;
+                    }
+                }
 
-                return seastar::make_ready_future<PartialUpdateResult>(PartialUpdateResult(std::move(status)));
+                return seastar::do_with( (k2::CachedSteadyClock::duration)sleep, [sleep=std::move(sleep),status=std::move(status)] (k2::CachedSteadyClock::duration& dur) {
```
iafuture, line 438: same issues as in the non-partial use case. You can probably just remove this for now since YCSB does only full-key updates, no partials right?

If not, you should try to refactor this a bit

src/k2/module/k23si/client/k23si_client.cpp
```diff
@@ -277,8 +280,8 @@ seastar::future<EndResult> K2TxnHandle::end(bool shouldCommit) {
             return _heartbeat_timer.stop().then([this, s=std::move(status)] () {
                 // TODO get min transaction time from TSO client
                 auto time_spent = Clock::now() - _start_time;
```
iafuture, line 282: we should check here if there is any need for sleep and only sleep if >0

src/k2/module/k23si/Module.cpp
```diff
@@ -1108,11 +1108,40 @@ K23SIPartitionModule::_processWrite(dto::K23SIWriteRequest&& request, FastDeadli
         }
     }
 
+    // build write_ranges for Transaction end request
+    std::unordered_map<String, std::unordered_set<dto::KeyRangeVersion>> write_ranges; // will be filled in later if WI is created successfully
```
iafuture, line 1112: no need for this

src/k2/module/k23si/Module.cpp
```diff
@@ -1108,11 +1108,40 @@ K23SIPartitionModule::_processWrite(dto::K23SIWriteRequest&& request, FastDeadli
...
+
+    //endRequest created for calling EndTransaction for one hot transactions.
+    dto::K23SITxnEndRequest endRequest{request.pvid,
+        request.trhCollection,
+        request.trh,
+        request.mtr,
+        dto::EndAction::Commit,
+        std::move(write_ranges)
```
iafuture, line 1120: just initialize as {}

src/k2/module/k23si/Module.cpp
```diff
@@ -1108,11 +1108,40 @@ K23SIPartitionModule::_processWrite(dto::K23SIWriteRequest&& request, FastDeadli
...
+    };
+
+    return seastar::do_with((dto::K23SITxnEndRequest)std::move(endRequest), (bool)request.isonehot,
```
iafuture, line 1123: no need for do-with. just move-capture in the lambda

src/k2/module/k23si/Module.cpp
```diff
@@ -1108,11 +1108,40 @@ K23SIPartitionModule::_processWrite(dto::K23SIWriteRequest&& request, FastDeadli
...
+                            [this, &request] (dto::K23SITxnEndRequest& endRequest, bool& isonehot) {
+        // all checks passed - we're ready to place this WI as the latest version
+        auto& vset = _indexer[request.key];
+        auto status = _createWI(std::move(request), vset);
+        K2LOG_D(log::skvsvr, "WI creation with status {}", status);
+
+        // not one hot transaction or WI creation failed
+        if(!isonehot || !status.is2xxOK()) {
+            return RPCResponse(std::move(status), dto::K23SIWriteResponse{});
+        }
```
iafuture, line 1133: you should create the end request above here only if needed i.e. WI succeeded and this is a 1-shot

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -269,52 +270,78 @@ class K2TxnHandle {
...
+            then([this, request=std::move(request), isonehot=std::move(isonehot)] (auto&& response) {
                 auto& [status, k2response] = response;
 
                 _registerRangeForWrite(status, *request);
 
                 _checkResponseStatus(status);
                 _ongoing_ops--;
 
-                if (status.is2xxOK() && !_heartbeat_timer.isArmed()) {
+                k2::CachedSteadyClock::duration sleep= 0us; // time to sleep before we return to satisfy min time_spent in txn
```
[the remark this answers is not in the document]
ammuv: done

src/k2/module/k23si/client/k23si_client.h
```diff
@@ -269,52 +270,78 @@ class K2TxnHandle {
...
+                    }
+                    //note we need not end heartbeat timer because we never started it for one hot transactions.
+                    auto time_spent = Clock::now() - _start_time;
+                    if (time_spent < 5us) {
+                        sleep = 5us - time_spent;
+                    }
+                }
 
-                return seastar::make_ready_future<WriteResult>(WriteResult(std::move(status), std::move(k2response)));
+                return seastar::do_with( (k2::CachedSteadyClock::duration)sleep, (dto::K23SIWriteResponse)std::move(k2response), [this,sleep=std::move(sleep),status=std::move(status)] (k2::CachedSteadyClock::duration& dur, dto::K23SIWriteResponse& k2response) {
```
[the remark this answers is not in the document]
ammuv: done

src/k2/module/k23si/client/k23si_client.h
ammuv: No YCSB does partial updates. A particular field is chosen at random and updated.

src/k2/module/k23si/client/k23si_client.cpp
ammuv: should raise a separate issue for this because this is part of the master branch also.

src/k2/module/k23si/Module.cpp
ammuv: done

src/k2/module/k23si/Module.cpp
ammuv: done

src/k2/module/k23si/Module.cpp
ammuv: done

src/k2/module/k23si/Module.cpp
ammuv: The problem here was that there is no copy constructor for write request and we need to move the request to create WI. To create an end request, we need the values in the write request, thus we need to create it before the write request is moved. I didn't want to add a copy constructor because it is not needed anywhere else.

Issue #211, opened 2021-08-12
Nodepool metrics


jfunston:
- committed txn rate
- read rate
- write rate
- query rate
- finalization rate
- active WIs
- active txns
- abort rate
- conflict abort rate

jfunston:
- persistence flush latency
- push rate
- push latency
- read, write, query processing latency
- finalization latency

jfunston:
- Number of keys in indexer
- Total number of record versions
- Total size of user data (the SKVRecord::Storage and payload size, and size of keys)

jfunston:
done

PR #215, opened 2021-08-27
Nodepool metrics

ammuv:
Added Nodepool metrics

Resolved issue #211


src/k2/module/k23si/Module.cpp
```diff
@@ -191,23 +191,57 @@ seastar::future<> K23SIPartitionModule::_registerVerbs() {
...
     RPC().registerRPCObserver<dto::K23SIReadRequest, dto::K23SIReadResponse>
     (dto::Verbs::K23SI_READ, [this](dto::K23SIReadRequest&& request) {
-        return handleRead(std::move(request), FastDeadline(_config.readTimeout()));
+        auto start = k2::Clock::now();
+        return handleRead(std::move(request), FastDeadline(_config.readTimeout()))
+               .then([this, start](auto&& response){
+                    _readOps++;
+                    auto end = k2::Clock::now();
+                    auto dur = end - start;
+                    _readLatency.add(dur);
```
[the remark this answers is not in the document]
iafuture: Also, I think i'd prefer a small utility class which can do the calculation and reporting, e.g. usage:
```c++
auto reporter = Reporter(_readLatency);
return handleRead(
   std::move(request),
   Fasteadline(_config.readTimeout()))
   .then([this, reporter=std::move(reporter)] (auto&& response) {
        _readOps++;
        _reporter.report();
        return std::move(response);
   }
```

We can then do something with compiler flags to compile the cost of  report() if we want to measure

src/k2/module/k23si/Module.cpp
```diff
@@ -284,7 +328,38 @@ void K23SIPartitionModule::_unregisterVerbs() {
...
+    _metric_groups.add_group("Nodepool", {
+        sm::make_gauge("indexer_keys",[this]{ return _indexer.size();},
+                        sm::description("Number of keys in indexer"), labels),
+        sm::make_counter("read_operations", _readOps, sm::description("Number of read operations"), labels),
+        sm::make_counter("write_operations", _writeOps, sm::description("Number of write operations"), labels),
+        sm::make_counter("query_operations", _queryOps, sm::description("Number of query operations"), labels),
+        sm::make_counter("finalization_operations", _finalizations, sm::description("Number of finalizations requests"), labels),
+        sm::make_counter("push_operations", _pushes, sm::description("Number of pushses"), labels),
+        sm::make_gauge("active_WI", _activeWI, sm::description("Number of active WIs"), labels),
+        sm::make_counter("record_versions", _recordVersions, sm::description("Number of record versions over all records"), labels),
```
iafuture, line 345: this should also be a gauge as it can go down (when we remove versions at end of retention window)

src/k2/module/k23si/Module.cpp
```diff
@@ -1129,6 +1204,7 @@ K23SIPartitionModule::_createWI(dto::K23SIWriteRequest&& request, VersionSet& ve
 
     // the TWIM accepted the write. Add it as a WI now
     versions.WI.emplace(std::move(rec), request.request_id);
+    _activeWI++;
```
iafuture, line 1207: I don't quite like the management of this count along with the std::optional. Is there some way we can tie the counter to the lifetime of the object?

src/k2/module/k23si/Persistence.cpp
```diff
@@ -65,11 +77,18 @@ seastar::future<> Persistence::stop() {
...
 
 seastar::future<Status> Persistence::flush() {
+    auto start = k2::Clock::now();
     ++_flushId;
     K2LOG_D(log::skvsvr, "flush with bs={}, proms={}, fid={}", (_buffer? _buffer->getSize() : 0), _pendingProms.size(), _flushId);
     if (!_buffer) {
         K2ASSERT(log::skvsvr, _pendingProms.size() == 0, "There is no data to send but we have pending promises");
-        return _chainFlushResponse();
+        return _chainFlushResponse()
+                .then([this, start](auto&& response){
```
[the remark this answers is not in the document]
iafuture: I think there may be a difference in interpreting what "flush" is. If we think as flush as "the duration it took us to flush out the buffered data", then it doesn't make sense to report here - we should only report the real flush once.

On the other hand if "flush" means "the time a user callback waited for flush to complete" then we should report here as we'd want to capture the train queue effect.

I think we should only measure as defined in $1. $2 is captured by the top-level write call

iafuture (changes requested):

src/k2/module/k23si/Persistence.cpp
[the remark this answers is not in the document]
ammuv: done

src/k2/module/k23si/Module.cpp
ammuv: done

src/k2/module/k23si/Module.cpp
ammuv: Discussed with Justin and we decided to use the _removeWI function in all places to maintain consistency. Just to clarify, the _activeWI keeps track of number of activeWI over all keys.

src/k2/module/k23si/Module.cpp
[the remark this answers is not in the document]
ammuv: done

src/k2/module/k23si/Module.cpp
```diff
@@ -191,23 +191,65 @@ seastar::future<> K23SIPartitionModule::_registerVerbs() {
 
     RPC().registerRPCObserver<dto::K23SIReadRequest, dto::K23SIReadResponse>
     (dto::Verbs::K23SI_READ, [this](dto::K23SIReadRequest&& request) {
-        return handleRead(std::move(request), FastDeadline(_config.readTimeout()));
+        OperationLatencyReporter reporter(_readOps, _readLatency);
+        #ifdef GET_LATENCY
+            reporter.startOperation();
```
jfunston, line 196: I don't think there should be a separate startOperation method, the work should be done in the constructor. Also maybe the compilation flag can be inside the Reporter class, and we can always count the number of ops even if the flag is off.

ammuv: Discussed offline with Justin. Decided to add extra compilation flag FORCE_GET_LATENCY when GET_LATENCY flag is not defined.

src/k2/module/k23si/Module.cpp
ammuv: Decided to introduce new metric to keep track of finalizedWI and change activeWI (gauge) metric to totalWI (counter).

ammuv: done

src/k2/module/k23si/Module.cpp
ammuv: done

jfunston (dismissed):

src/k2/module/k23si/utils.h
```diff
@@ -0,0 +1,52 @@
...
+        OperationLatencyReporter(uint64_t& ops, k2::ExponentialHistogram& latencyHist): _ops(ops), _latencyHist(latencyHist){
+            #if defined GET_LATENCY || defined FORCE_GET_LATENCY
+                _startTime = k2::Clock::now(); // record start time to calculate latency of operation
+            #endif
+        }
+
+        // update operation count and latency metrics
+        void report(){
+            _ops++;
+            #if defined GET_LATENCY || defined FORCE_GET_LATENCY
```
iafuture, line 39: this conditional compilation won't work unless these are defined right before the file is included.

In other words, you can only use this if in the reporting file you do:
```c++
#define GET_LATENCY
#include <utils.h>
....
```

and undefining it doesn't make a difference in the including file

src/k2/module/k23si/utils.h
```diff
@@ -0,0 +1,52 @@
+/*
```
iafuture, line 1: The contents of this file are metric-specific. We have some metric stuff in the Prometheus.h header in transport. please place this there.

In general, try not to use general names such as "utils" or "common" for stuff. Conversely, if you should come across some code in a common/utils file and you have an idea of how to name it better, do feel free to make the change :)

iafuture (changes requested):

src/k2/module/k23si/utils.h
ammuv: done

src/k2/module/k23si/utils.h
ammuv: Done. Resolved using config variable as per discussion.

src/k2/module/k23si/Persistence.cpp
```diff
@@ -65,6 +77,7 @@ seastar::future<> Persistence::stop() {
 }
 
 seastar::future<Status> Persistence::flush() {
+    auto start = k2::Clock::now();
```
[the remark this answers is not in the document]
ammuv: Done.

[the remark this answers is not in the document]
ammuv: Added metric for number of flushes as well.

iafuture (approved):

PR #216, opened 2021-08-31
Fix unsafe capture in TPCC transactions

### LINKED ISSUE #191: Unsafe capture in getCIdByLastName in transactions.h
lastName capture by reference is unsafe.

jfunston:
closes #191


jfunston:
Also make TPC-C randomness for last name generation match spec more closely

iafuture (approved):

PR #222, opened 2021-09-11
change to use BinaryAllocator to replace BinaryAllocatorFunctor

johnfangAFW:
Change to use a BinaryAllocator to replace the original BinaryAllocatorFunctor due to the following reasons:
1) holds a default block size, which the function could not have such a state
2) provide a method to allocate by a specific size instead of the default block size

All unit tests and integration tests passed.

root@b0d49a77c3ce:/build# cd build/test && ctest
Test project /build/build/test
    Start 1: plogmock
1/9 Test #1: plogmock .........................   Passed    0.38 sec
    Start 2: persistent_volume
2/9 Test #2: persistent_volume ................   Passed    3.09 sec
    Start 3: transport
3/9 Test #3: transport ........................   Passed    0.09 sec
    Start 4: readcache
4/9 Test #4: readcache ........................   Passed    0.03 sec
    Start 5: skv_record
5/9 Test #5: skv_record .......................   Passed    0.05 sec
    Start 6: key_encoding
6/9 Test #6: key_encoding .....................   Passed    0.04 sec
    Start 7: skv_ser
7/9 Test #7: skv_ser ..........................   Passed    0.04 sec
    Start 8: expression
8/9 Test #8: expression .......................   Passed    0.05 sec
    Start 9: transport
9/9 Test #9: transport ........................   Passed    0.04 sec

100% tests passed, 0 tests failed out of 9

Total Test time (real) =   3.96 sec

### DIFF
--- a/src/k2/transport/Payload.cpp
+++ b/src/k2/transport/Payload.cpp
@@ -36,7 +36,7 @@ Payload::PayloadPosition::PayloadPosition(_Size bIdx, _Size bOff, size_t offset)
     // Empty payload will have position of (bidx=0, boff=0, off=0)
 }
 
-Payload::Payload(seastar::lw_shared_ptr<BinaryAllocator> allocator):
+Payload::Payload(BinaryAllocator allocator):
     _size(0), _capacity(0), _allocator(allocator) {
 }
 
@@ -44,17 +44,17 @@ Payload::Payload(std::vector<Binary>&& externallyAllocatedBuffers, size_t contai
     _buffers(std::move(externallyAllocatedBuffers)),
     _size(containedDataSize),
     _capacity(_size),
-    _allocator(nullptr) {
+    _allocator(BinaryAllocator()) {
 }
 
 Payload::Payload():
     _size(0),
     _capacity(0),
-    _allocator(nullptr) {
+    _allocator(BinaryAllocator()) {
 }
 
-seastar::lw_shared_ptr<BinaryAllocator> Payload::DefaultAllocator(size_t default_size) {
-    return seastar::make_lw_shared<BinaryAllocator>(default_size, [](size_t bsize) { return Binary(bsize); });
+BinaryAllocator Payload::DefaultAllocator(size_t default_size) {
+    return BinaryAllocator(default_size, [](size_t bsize) { return Binary(bsize); });
 }
 
 bool Payload::isEmpty() const {
@@ -93,7 +93,7 @@ void Payload::clear() {
 
 void Payload::appendBinary(Binary&& binary) {
     // we can only append into a non-self-allocating payload
-    K2ASSERT(log::tx, _allocator.get() == nullptr, "cannot append to non-allocating payload");
+    K2ASSERT(log::tx, !_allocator.canAllocate(), "cannot append to non-allocating payload");
     _size += binary.size();
     _capacity += binary.size();
     _buffers.push_back(std::move(binary));
@@ -201,7 +201,7 @@ bool Payload::read(Payload& other) {
     other.clear();
     other._size = size;
     other._capacity = size;
-    other._allocator = nullptr;
+    other._allocator = BinaryAllocator();
     other._currentPosition = PayloadPosition();
     while(size > 0) {
         auto shared = _buffers[_currentPosition.bufferIndex].share();
@@ -333,8 +333,8 @@ size_t Payload::getFieldsSize() {
 }
 
 bool Payload::_allocateBuffer() {
-    K2ASSERT(log::tx, _allocator, "cannot allocate buffer without allocator");
-    Binary buf = _allocator->allocate();
+    K2ASSERT(log::tx, _allocator.canAllocate(), "cannot allocate buffer without allocator");
+    Binary buf = _allocator.allocate();
     if (!buf) {
         return false;
     }
@@ -434,9 +434,9 @@ Payload Payload::shareRegion(size_t startOffset, size_t nbytes){
     return shared;
 }
 
-Payload Payload::copy(seastar::lw_shared_ptr<BinaryAllocator> allocator) {
+Payload Payload::copy(BinaryAllocator allocator) {
     Payload copied;
-    Binary b = allocator->allocate(_size);
+    Binary b = allocator.allocate(_size);
 
     // copy exactly the data we need
     size_t toCopy = _size;
--- a/src/k2/transport/Payload.cpp
+++ b/src/k2/transport/Payload.cpp
@@ -36,8 +36,8 @@ Payload::PayloadPosition::PayloadPosition(_Size bIdx, _Size bOff, size_t offset)
     // Empty payload will have position of (bidx=0, boff=0, off=0)
 }
 
-Payload::Payload(BinaryAllocator allocator):
-    _size(0), _capacity(0), _allocator(allocator) {
+Payload::Payload(BinaryAllocator&& allocator):
+    _size(0), _capacity(0), _allocator(std::move(allocator)) {
 }
 
 Payload::Payload(std::vector<Binary>&& externallyAllocatedBuffers, size_t containedDataSize):
@@ -409,7 +409,8 @@ Payload Payload::shareRegion(size_t startOffset, size_t nbytes){
     seek(startOffset);
     nbytes = std::min(_size - _currentPosition.offset, nbytes);
 
-    Payload shared(_allocator);
+    BinaryAllocator copied_allocator = _allocator;
+    Payload shared(std::move(copied_allocator));
     shared._size = nbytes;
     shared._capacity = nbytes; // the capacity of the new payload stops with the current data written
 
@@ -434,7 +435,7 @@ Payload Payload::shareRegion(size_t startOffset, size_t nbytes){
     return shared;
 }
 
-Payload Payload::copy(BinaryAllocator allocator) {
+Payload Payload::copy(BinaryAllocator&& allocator) {
     Payload copied;
     Binary b = allocator.allocate(_size);
 
@@ -451,7 +452,7 @@ Payload Payload::copy(BinaryAllocator allocator) {
     }
 
     copied.appendBinary(std::move(b));
-    copied._allocator = allocator;
+    copied._allocator = std::move(allocator);
     return copied;
 }
 
--- a/src/k2/transport/Payload.cpp
+++ b/src/k2/transport/Payload.cpp
@@ -53,8 +53,8 @@ Payload::Payload():
     _allocator(nullptr) {
 }
 
-std::shared_ptr<BinaryAllocator> Payload::DefaultAllocator() {
-    return std::make_shared<BinaryAllocator>(8196);
+std::shared_ptr<BinaryAllocator> Payload::DefaultAllocator(size_t default_size) {
+    return std::make_shared<BinaryAllocator>(default_size, [](size_t bsize) { return Binary(bsize); });
 }
 
 bool Payload::isEmpty() const {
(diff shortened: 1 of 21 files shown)


iafuture (changes requested):
The proposed solution here does solve the problem but I think we should keep the lambda-based approach as it is more efficient and quite a bit more flexible. 

The efficiency issue is mostly due to the now required allocations from the shared_ptr usage. You can avoid that by accepting a BinaryAllocator instance instead.

The bigger issue though is that this approach is not as flexible. The user of Payload now only gets to specify what size blocks are allocated but not how or where they come from. With a lambda, I can have you allocate memory from a designated area, such as SCM, or DPDK-marked memory, or even from a different thread. To achieve the same with the class approach, you'll have to now build inheritance-based solution which gets things even more inefficient due to vtables.

I'd revert back to the lambda-based approach with a default size in the TCP and RDMA implementations(VirtualNetworkStack.cpp), as we discussed in the online chat. Let's do another chat if you have more questions

johnfangAFW:
> The proposed solution here does solve the problem but I think we should keep the lambda-based approach as it is more efficient and quite a bit more flexible.
> 
> The efficiency issue is mostly due to the now required allocations from the shared_ptr usage. You can avoid that by accepting a BinaryAllocator instance instead.
> 
> The bigger issue though is that this approach is not as flexible. The user of Payload now only gets to specify what size blocks are allocated but not how or where they come from. With a lambda, I can have you allocate memory from a designated area, such as SCM, or DPDK-marked memory, or even from a different thread. To achieve the same with the class approach, you'll have to now build inheritance-based solution which gets things even more inefficient due to vtables.
> 
> I'd revert back to the lambda-based approach with a default size in the TCP and RDMA implementations(VirtualNetworkStack.cpp), as we discussed in the online chat. Let's do another chat if you have more questions

As discussed, changed to pass both default size and the std::function to BinaryAllocator to avoid inheritance.

src/k2/transport/Payload.cpp
```diff
@@ -36,7 +36,7 @@ Payload::PayloadPosition::PayloadPosition(_Size bIdx, _Size bOff, size_t offset)
     // Empty payload will have position of (bidx=0, boff=0, off=0)
 }
 
-Payload::Payload(BinaryAllocatorFunctor allocator):
+Payload::Payload(std::shared_ptr<BinaryAllocator> allocator):
```
jfunston, line 39: Can we pass the allocator by move instead of a shared pointer?

jfunston (changes requested):
Maybe you can also run the test_k23si_tpcc.sh test locally and see if there is any noticeable performance degradation, even though I don't expect there to be.

src/k2/transport/Payload.cpp
johnfangAFW: I chose to use a shared pointer here to give users the flexibility to either use an individual allocator for a payload or use a shared allocator for multiple payloads. If we have the use case to use the same allocator for many payloads, the shared pointer is more efficient so that we don't have to create an allocator instance for each payload.

Run test_k23si_tpcc.sh locally and compare it with the master, didn't see any performance degradation. The times are almost the same.

jfunston: I see. If we want to keep the shared pointer we can change it to a seastar::lw_shared_ptr which is faster. The restrictions on it are that there is no polymorphism and it stays on the same core, which we should meet for this use case.

johnfangAFW: I see, will change to use seastar::lw_shared_ptr then.

johnfangAFW: done

src/k2/transport/Payload.cpp
```diff
@@ -36,7 +36,7 @@ Payload::PayloadPosition::PayloadPosition(_Size bIdx, _Size bOff, size_t offset)
     // Empty payload will have position of (bidx=0, boff=0, off=0)
 }
 
-Payload::Payload(BinaryAllocatorFunctor allocator):
+Payload::Payload(seastar::lw_shared_ptr<BinaryAllocator> allocator):
```
iafuture, line 39: I think you should use allocator objects instead of shared pointer. The shared_ptr approach is more expensive since it requires allocation

iafuture (changes requested):

src/k2/transport/Payload.cpp
johnfangAFW: The purpose of a shared pointer here is to allow multiple payloads to use the same allocator to avoid allocating a BinaryAllocator for each individual payload. It is more expensive to use an allocator object for each payload in such a use case.

iafuture: I think the shared_ptr is more expensive. A shared pointer copies a pointer and increments a shared counter. The functor object on the other hand is just a copy of a pointer (the function underneath).

Moreover, the shared pointer does require allocation, as you do in the rest of the code base (calling make_lw_shared). You could "share" the shared_ptr, but this is just as expensive since just copying the object since it is so light.
Lastly, the shared_ptr approach also now requires 2x dereferencing (1 for the shared_ptr, 2 for the function in the object) whereas with the object it is just 1.

johnfangAFW: I see, this is because the BinaryAllocator object is so light that the direct object copy here is less expensive than copying a shared pointer and updating the object reference count in the shared pointer.

Will change to use the object directly.

johnfangAFW: done

src/k2/transport/VirtualNetworkStack.cpp
```diff
@@ -24,6 +24,7 @@ Copyright(c) 2020 Futurewei Cloud
 #include "VirtualNetworkStack.h"
 
 // third-party
+#include <memory>
```
iafuture, line 27: probably don't need this

src/k2/transport/TXEndpoint.cpp
```diff
@@ -72,18 +72,18 @@ std::unique_ptr<TXEndpoint> TXEndpoint::fromURL(const String& url, BinaryAllocat
...
     K2LOG_D(log::tx, "dtor");
 }
 
-TXEndpoint::TXEndpoint(String&& pprotocol, String&& pip, uint32_t pport, BinaryAllocatorFunctor&& allocator):
+TXEndpoint::TXEndpoint(String&& pprotocol, String&& pip, uint32_t pport, BinaryAllocator allocator):
     protocol(std::move(pprotocol)),
     ip(std::move(pip)),
     port(pport),
-    _allocator(std::move(allocator)) {
+    _allocator(allocator) {
```
iafuture, line 86: you should still move here

src/k2/transport/TXEndpoint.cpp
```diff
@@ -72,18 +72,18 @@ std::unique_ptr<TXEndpoint> TXEndpoint::fromURL(const String& url, BinaryAllocat
...
-    return std::make_unique<TXEndpoint>(std::move(protocol), std::move(ip), port, std::move(allocator));
+    return std::make_unique<TXEndpoint>(std::move(protocol), std::move(ip), port, allocator);
 }
 
 TXEndpoint::~TXEndpoint() {
     K2LOG_D(log::tx, "dtor");
 }
 
-TXEndpoint::TXEndpoint(String&& pprotocol, String&& pip, uint32_t pport, BinaryAllocatorFunctor&& allocator):
+TXEndpoint::TXEndpoint(String&& pprotocol, String&& pip, uint32_t pport, BinaryAllocator allocator):
```
iafuture, line 82: receive the allocator by && to avoid a copy

src/k2/transport/TXEndpoint.h
```diff
@@ -44,7 +44,7 @@ class TXEndpoint {
 public: // lifecycle
     // construct an endpoint from a url with the given allocator
     // Returns nullptr if there was a problem parsing the url
-    static std::unique_ptr<TXEndpoint> fromURL(const String& url, BinaryAllocatorFunctor&& allocator);
+    static std::unique_ptr<TXEndpoint> fromURL(const String& url, BinaryAllocator allocator);
```
iafuture, line 47: why this change? you should still receive by && to avoid a copy

src/k2/transport/TXEndpoint.cpp
```diff
@@ -38,7 +38,7 @@ namespace k2 {
 // must have: protocol(group1), ip(group2==ipv4, group3==ipv6), port(group4)
 const std::regex urlregex{"(.+)://(?:([^:\\[\\]]+)|\\[(.+)\\]):(\\d+)"};
 
-std::unique_ptr<TXEndpoint> TXEndpoint::fromURL(const String& url, BinaryAllocatorFunctor&& allocator) {
+std::unique_ptr<TXEndpoint> TXEndpoint::fromURL(const String& url, BinaryAllocator allocator) {
```
iafuture, line 41: receive by && to avoid copy

src/k2/transport/Payload.h
```diff
@@ -165,7 +165,10 @@ class Payload {
     Payload shareRegion(size_t startOffset, size_t nbytes);
 
     // Creates a new payload as a copy of this payload. The underlying data is copied over to the new payload
-    Payload copy(BinaryAllocatorFunctor dest_allocator=DefaultAllocator);
+    Payload copy(BinaryAllocator dest_allocator);
```
iafuture, line 168: since we're keeping this allocator, i think you should receive by &&

src/k2/transport/Payload.cpp
```diff
@@ -434,8 +434,9 @@ Payload Payload::shareRegion(size_t startOffset, size_t nbytes){
     return shared;
 }
 
-Payload Payload::copy(BinaryAllocatorFunctor allocator) {
-    Payload copied(allocator);
+Payload Payload::copy(BinaryAllocator allocator) {
```
iafuture, line 437: receive by && and then move later on into copied._allocator

iafuture (changes requested):

src/k2/transport/VirtualNetworkStack.cpp
johnfangAFW: removed

src/k2/transport/TXEndpoint.cpp
johnfangAFW: done. But we need to manually copy the allocator in some object for a shared allocator to avoid making the original allocator as empty.

src/k2/transport/TXEndpoint.h
johnfangAFW: done

src/k2/transport/Payload.h
johnfangAFW: done

src/k2/transport/Payload.cpp
johnfangAFW: done, but we need to make a copy of a shared allocator in some cases to avoid making the original allocator empty/invalid.

src/k2/transport/TXEndpoint.cpp
johnfangAFW: done

src/k2/transport/TXEndpoint.cpp
johnfangAFW: done

jfunston (dismissed):

src/k2/transport/Payload.cpp
```diff
@@ -409,7 +409,8 @@ Payload Payload::shareRegion(size_t startOffset, size_t nbytes){
     seek(startOffset);
     nbytes = std::min(_size - _currentPosition.offset, nbytes);
 
-    Payload shared(_allocator);
+    BinaryAllocator copied_allocator = _allocator;
+    Payload shared(std::move(copied_allocator));
```
iafuture, line 413: I think this could just be
```suggestion
    Payload shared(BinaryAllocator(_allocator));
```

johnfangAFW: done, thanks for suggestions.

iafuture (approved):

PR #223, opened 2021-09-13
Query Metrics

ammuv:
Added histograms for query scans and returns.

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -331,7 +331,11 @@ void K23SIPartitionModule::_registerMetrics() {
         sm::make_histogram("query_latency", [this]{ return _queryLatency.getHistogram();},
                 sm::description("Latency of Query Operations"), labels),
         sm::make_histogram("push_latency", [this]{ return _pushLatency.getHistogram();},
-                sm::description("Latency of Pushes"), labels)
+                sm::description("Latency of Pushes"), labels),
+        sm::make_histogram("query_scans", [this]{ return _queryScans.getHistogram();},
+                sm::description("Number of records scanned by query operations"), labels),
+        sm::make_histogram("query_returns", [this]{ return _queryReturns.getHistogram();},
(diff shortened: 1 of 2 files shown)


src/k2/module/k23si/Module.cpp
```diff
@@ -585,7 +585,7 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
 
     IndexerIterator key_it = _initializeScan(request.key, request.reverseDirection, request.exclusiveKey);
     for (; !_isScanDone(key_it, request, response.results.size());
-                        _scanAdvance(key_it, request.reverseDirection, request.key.schemaName)) {
+                        _scanAdvance(key_it, request.reverseDirection, request.key.schemaName), numScans++) {
```
jfunston, line 588: I think it is better to move the increment into the for loop body, it is easy to miss here

src/k2/module/k23si/Module.cpp
```diff
@@ -646,12 +646,12 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
         request.key = key_it->first; // if we retry, do so with the key we're currently iterating on
         return _doPush(request.key, versions.WI->data.timestamp, request.mtr, deadline)
```
jfunston, line 647: I think maybe we shouldn't track numScans across pushes. The idea is to get a feel for how much the reactor is monopolized by the query, and the push will yield it. We can see what Ivan thinks.

jfunston (changes requested):

src/k2/module/k23si/Module.cpp
ammuv: done.

src/k2/module/k23si/Module.cpp
ammuv: done.

jfunston (dismissed):

iafuture (approved):

PR #229, opened 2021-09-24
ycsb tpcc retry change and ycsb szipfian overflow change

jfunston:


src/k2/cmd/ycsb/ycsb_rand.h
```diff
@@ -199,16 +199,17 @@ uint64_t fnvhash64(uint64_t val) {
     static const int64_t FNV_PRIME_64 = 1099511628211L;
 
     //from http://en.wikipedia.org/wiki/Fowler_Noll_Vo_hash
-    int64_t hashval = FNV_OFFSET_BASIS_64;
+    uint64_t hashval = FNV_OFFSET_BASIS_64;
+    uint64_t octet;
```
iafuture, line 203: This octet var here is confusing. It should be just declared inline with assignment on line 206

iafuture (changes requested):

iafuture (approved):

Issue #241, opened 2021-10-12
TPCC load crash issue


johnfangAFW:
tpcc_client: /build/src/core/future.cc:150: void seastar::future_state_base::ignore(): Assertion `0 && "invalid state for ignore"' failed.
Aborting on shard 4.
Backtrace:
0x00000000003387ec
0x00000000002fff60
0x0000000000300262
0x0000000000300322
0x00002a34acdfb3bf
/lib/x86_64-linux-gnu/libc.so.6+0x000000000004618a
/lib/x86_64-linux-gnu/libc.so.6+0x0000000000025858
/lib/x86_64-linux-gnu/libc.so.6+0x0000000000025728
/lib/x86_64-linux-gnu/libc.so.6+0x0000000000036f35
0x00000000002ca4fb
0x000000000014540f
0x00000000001455d7
0x00000000002faa97
0x00000000002fade8
0x0000000000324ddd
0x000000000032b7bb
0x00000000002e07c1
/lib/x86_64-linux-gnu/libpthread.so.0+0x0000000000009608
/lib/x86_64-linux-gnu/libc.so.6+0x0000000000122292

johnfangAFW:
# echo "0x00000000003387ec
> 0x00000000002fff60
> 0x0000000000300262
> 0x0000000000300322
> 0x00002a34acdfb3bf
> /lib/x86_64-linux-gnu/libc.so.6+0x000000000004618a
> /lib/x86_64-linux-gnu/libc.so.6+0x0000000000025858
> /lib/x86_64-linux-gnu/libc.so.6+0x0000000000025728
> /lib/x86_64-linux-gnu/libc.so.6+0x0000000000036f35
> 0x00000000002ca4fb
> 0x000000000014540f
> 0x00000000001455d7
> 0x00000000002faa97
> 0x00000000002fade8
> 0x0000000000324ddd
> 0x000000000032b7bb
> 0x00000000002e07c1
> /lib/x86_64-linux-gnu/libpthread.so.0+0x0000000000009608
> /lib/x86_64-linux-gnu/libc.so.6+0x0000000000122292" | addr2line -e ./build/src/k2/cmd/tpcc/tpcc_client
/build/include/seastar/util/backtrace.hh:59
/build/src/core/reactor.cc:732
/build/src/core/reactor.cc:760
/build/src/core/reactor.cc:3550
??:0
??:0
??:0
??:0
??:0
/build/src/core/future.cc:150
[...]

PR #243, opened 2021-10-15
fix tpcc verification issues

johnfangAFW:
Fixed TPCC verification failures due to invalid future states and deadline limit in local machine.

Analysis is in:
https://github.com/futurewei-cloud/chogori-platform/issues/241

After the fix, both unit tests and TPCC tests passed.

### DIFF
--- a/src/k2/cmd/tpcc/schema.h
+++ b/src/k2/cmd/tpcc/schema.h
@@ -39,7 +39,7 @@ static const String tpccCollectionName = "TPCC";
 #define CHECK_READ_STATUS(read_result) \
     do { \
         if (!((read_result).status.is2xxOK())) { \
-            K2LOG_D(log::tpcc, "TPC-C failed to read rows: {}", (read_result).status); \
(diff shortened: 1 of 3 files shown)


src/k2/cmd/tpcc/schema.h
```diff
@@ -39,7 +39,7 @@ static const String tpccCollectionName = "TPCC";
 #define CHECK_READ_STATUS(read_result) \
     do { \
         if (!((read_result).status.is2xxOK())) { \
-            K2LOG_D(log::tpcc, "TPC-C failed to read rows: {}", (read_result).status); \
+            K2LOG_E(log::tpcc, "TPC-C failed to read rows: {}", (read_result).status); \
```
jfunston, line 42: We should keep these and the three other log lines in the helper functions as debug because they can be normal expected failures during the benchmark (e.g. conflict abort)

johnfangAFW: I see. Reverted back.

jfunston (approved):
Looks good, thanks!

PR #248, opened 2021-11-02
update getTimeNow() in test to return increasing timestamps

johnfangAFW:
Fix the reported issue that the  getTimeNow() in test could return the same timestamps in some scenarios. The original idea to use initial time and offset had some issues, for example, 

[0000:00:04:55.167.190]-3si_txn_test-[0]-(k2::k23si_test) [ERROR] [/build/test/k23si/3SITxnTest.cpp:1159 @operator()] status({code=403, message=TRH create request is too old}) == dto::K23SIStatus::Created({code=201, message=})
3si_txn_test: /build/test/k23si/3SITxnTest.cpp:1159: k2::txn_testing::testScenario01()::<lambda(k2::dto::Timestamp&&)>::<lambda(k2::dto::K23SI_MTR&, k2::dto::Key&, k2::dto::Key&, k2::dto::Key&, k2::DataRec&, k2::DataRec&)>::<lambda(auto:122&&)> [with auto:122 = std::tuple<k2::Status, k2::dto::K23SIWriteResponse>]: Assertion `((status) == (dto::K23SIStatus::Created))' failed.

As a result, add the logic to check for the generate timestamp and sleep if necessary to make sure that we return an increasing timestamp.

All tests passed.

### DIFF
--- a/test/include/TestUtil.h
+++ b/test/include/TestUtil.h
@@ -23,7 +23,9 @@ Copyright(c) 2020 Futurewei Cloud
 
 #pragma once
 #include <k2/common/Common.h>
+#include <k2/common/Chrono.h>
 #include <k2/common/Log.h>
+#include <k2/dto/Timestamp.h>
 #include <pthread.h>
 #include <filesystem>
 #include <iostream>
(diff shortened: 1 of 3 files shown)


johnfangAFW:
TSO work will cover this test. discard this for now.

PR #249, opened 2021-11-09
Tso revise

iafuture:
- Enabled all integration tests, including tpcc/ycsb
- Remove batching in TSO client and servers
- Add GPSClock provider
- assign TSO_ID per worker
- add API to obtain a timestamp from TSO
- consistent naming for cpo/tso service and client
- change integration tests to talk to TSO service
- Streamline Chrono API
- skv client txn duration now starts at time of receiving TSO timestamp
- RetryStrategy users can now signal to stop retries via exception
- Add metrics to TSO client
- fix WI metric in module


src/k2/tso/service/Clock.h
```diff
@@ -0,0 +1,152 @@
...
+
+private:
+    // obtain gps and steady timestamps
+    static uint64_t _gpsNow();
+    static uint64_t _steadyNow();
+    // Holds the same information as a GPSTimePoint, but in a volatile-friendly data members
+    struct _GPSNanos {
+        _GPSNanos();
+        _GPSNanos(uint64_t steady, uint64_t real, uint32_t error);
+        void operator=(const volatile _GPSNanos& o) volatile;
```
jfunston, line 99: Do we really want these methods declared volatile? My understanding is that it will treat all member variables as volatile within the functions. I'm not sure we always want that.

src/k2/tso/service/Service.cpp
```diff
@@ -0,0 +1,220 @@
...
+        K2LOG_I(log::tsoserver, "TSOService clock has been initialized with GPS steady={}, real={}, error={}",
+                     nsec(now.steady).count(), nsec(now.real).count(), nsec(now.error).count());
+        if (now.error > _errorBound()) {
+            K2LOG_E(log::tsoserver, "TSOService cannot start since the GPS error is larger than our error bound {}", _errorBound());
+            return seastar::make_exception_future<>(std::runtime_error("gps error is bigger than max error bound"));
+        }
+        // start the poller thread
+        K2LOG_I(log::tsoserver, "Starting GPS clock poller on CPU: {}", _clockPollerCPU());
+        _keepRunningPoller.test_and_set(); // set the running flag to true
+        _clockPoller = std::thread([this, pinCPU=_clockPollerCPU()] {
```
jfunston, line 61: maybe just capture _keepRunningPoller by reference instead of "this"?

src/k2/tso/service/Service.cpp
```diff
@@ -0,0 +1,220 @@
...
+seastar::future<> TSOService::_cpoRegister() {
+    // TODO implement CPO-based registration and ID assignment
+    _tsoId = seastar::this_shard_id() + 100000;
+
+    K2LOG_I(log::tsoserver, "Registration with CPO successful with id {}", _tsoId);
+    return seastar::make_ready_future<>();
+}
+
+seastar::future<std::tuple<Status, dto::GetServiceNodeURLsResponse>>
+TSOService::_handleGetServiceNodeURLs(dto::GetServiceNodeURLsRequest&&) {
```
jfunston, line 146: What is the purpose of this RPC? Is it just to share the RDMA endpoint? It seems that you need to have the worker url to call this to begin with.

src/k2/tso/service/Service.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // FYI, probably the GPS timestamp did not update across consecutive calls to TSO.
+    if (endCount <= _lastGeneratedEndCount) {
+        // increment the last returned value and use it as the end count
+        ++_lastGeneratedEndCount;
+        endCount = _lastGeneratedEndCount;
+    }
+
+    // ensure the result is a valid TSO timestamp within error bound
+    if ((uint64_t)nsec(_lastGPSTime.real - _lastGPSTime.error).count() < endCount - delta ||
+        (uint64_t)nsec(_lastGPSTime.real + _lastGPSTime.error).count() > endCount) {
```
jfunston, line 200: Shouldn't this condition always be true if your check above (_lastGPSTime.error > _errorBound()/2) passes?

jfunston (changes requested):

src/k2/tso/client/Client.cpp
```diff
@@ -0,0 +1,224 @@
...
+        .then([this](auto&& response) {
+            auto& [status, r] = response;
+            if (!status.is2xxOK()) {
+                K2LOG_E(log::tsoclient, "Error during get TSO node URLs, status:{}", status);
+                if (status.is5xxRetryable()) {
+                    // retryable errors
+                    return seastar::make_exception_future<>(std::runtime_error(status.message));
+                }
+                else {
+                    return seastar::make_exception_future<>(StopRetryException{});
```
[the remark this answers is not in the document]
iafuture: i can comment to mention this should only be called from a retryStrategy

src/k2/tso/service/Service.cpp
```diff
@@ -0,0 +1,220 @@
...
+            // pin the calling thread to the given CPU
+            if (pinCPU >= 0) {
+                cpu_set_t cpuset;
+                CPU_ZERO(&cpuset);
+                CPU_SET(pinCPU, &cpuset);
+                if (0 != pthread_setaffinity_np(pthread_self(), sizeof(cpu_set_t), &cpuset)) {
+                    throw std::runtime_error("Unable to set affinity");
+                }
+            }
+            while (_keepRunningPoller.test_and_set()) {
```
johnfangAFW, line 71: why do we need to call test_and_set() to set the flag to true here? shouldn't we just check if the flag is true and stop polling when the flag is false?

test/integration/failing_test_heartbeat.sh
```diff
@@ -5,25 +5,28 @@ set -e
...
 
-# start CPO on 1 cores
+# start CPO
 ./build/src/k2/cmd/controlPlaneOracle/cpo_main -c1 --tcp_endpoints ${CPO} --data_dir ${CPODIR} --enable_tx_checksum true --reactor-backend epoll --thread-affinity false --prometheus_port 63000 --assignment_timeout=1s --heartbeat_deadline=1s &
 cpo_child_pid=$!
 
-# start tso on 2 cores
-./build/src/k2/cmd/tso/tso -c2 --tcp_endpoints ${TSO} 13001 --enable_tx_checksum true --reactor-backend epoll --prometheus_port 63003 --thread-affinity false &
+# start tso
+./build/src/k2/cmd/tso/tso -c1 --tcp_endpoints ${TSO} --enable_tx_checksum true --reactor-backend epoll --prometheus_port 63003 --thread-affinity false --tso.error_bound=100us --tso.clock_poller_cpu=${TSO_POLLER_CORE} &
```
johnfangAFW, line 22: Is the 1 core, i.e., -c1, only for the poller? how do we specify the number of workers on different cores?

jfunston (approved):

PR #251, opened 2021-11-16
do not allow infinite push for read and write operations

johnfangAFW:
The bug was found by Jepsen tests. This PR is to only allow the push count for a key to be at most k23si_max_push_count, which is 1 by default.

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -199,7 +199,7 @@ seastar::future<> K23SIPartitionModule::_registerVerbs() {
         }
 
         k2::OperationLatencyReporter reporter(_readLatency); // for reporting metrics
-        return handleRead(std::move(request), FastDeadline(_config.readTimeout()))
+        return handleRead(std::move(request), FastDeadline(_config.readTimeout()), 0)
                .then([this, reporter=std::move(reporter)](auto&& response) mutable {
                     reporter.report();
                     return std::move(response);
@@ -213,7 +213,7 @@ seastar::future<> K23SIPartitionModule::_registerVerbs() {
         }
 
         k2::OperationLatencyReporter reporter(_queryPageLatency); // for reporting metrics
-        return handleQuery(std::move(request), dto::K23SIQueryResponse{}, FastDeadline(_config.readTimeout()))
+        return handleQuery(std::move(request), dto::K23SIQueryResponse{}, FastDeadline(_config.readTimeout()), 0)
                 .then([this, reporter=std::move(reporter)] (auto&& response) mutable {
                     reporter.report();
                     return std::move(response);
@@ -601,7 +601,7 @@ std::tuple<Status, bool> K23SIPartitionModule::_doQueryFilter(dto::K23SIQueryReq
 }
 
 seastar::future<std::tuple<Status, dto::K23SIQueryResponse>>
-K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQueryResponse&& response, FastDeadline deadline) {
+K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQueryResponse&& response, FastDeadline deadline, uint32_t count) {
     K2LOG_D(log::skvsvr, "Partition: {}, received query {}", _partition, request);
 
     uint64_t numScans = 0;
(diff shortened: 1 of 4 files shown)


johnfangAFW:
// check to see if we should push or is this a write from same txn

    if (vset.WI.has_value() && vset.WI->data.timestamp != request.mtr.timestamp) {

        // this is a write request finding a WI from a different transaction. Do a push with the remaining

        // deadline time.

        K2LOG_D(log::skvsvr, "different WI found for key {}", request.key);

        return _doPush(request.key, vset.WI->data.timestamp, request.mtr, deadline)

            .then([this, request = std::move(request), deadline](auto&& retryChallenger) mutable {

                if (!retryChallenger.is2xxOK()) {

                    // challenger must fail. Flush in case a TR was created during this call to handle write

                    K2LOG_D(log::skvsvr, "write push challenger lost for key {}", request.key);

                    return RPCResponse(dto::K23SIStatus::AbortConflict("incumbent txn won in write push"), dto::K23SIWriteResponse{});

                }



                K2LOG_D(log::skvsvr, "write push retry for key {}", request.key);

                return _processWrite(std::move(request), deadline);            —> line 1059

            });

    }

jfunston (dismissed):

src/k2/module/k23si/Module.cpp
```diff
@@ -749,13 +754,17 @@ K23SIPartitionModule::handleRead(dto::K23SIReadRequest&& request, FastDeadline d
...
     // record is still pending and isn't from same transaction.
     return _doPush(request.key, versions.WI->data.timestamp, request.mtr, deadline)
-        .then([this, request=std::move(request), deadline](auto&& retryChallenger) mutable {
+        .then([this, request=std::move(request), deadline, count](auto&& retryChallenger) mutable {
             if (!retryChallenger.is2xxOK()) {
                 return RPCResponse(dto::K23SIStatus::AbortConflict("incumbent txn won in read push"), dto::K23SIReadResponse{});
             }
-            return handleRead(std::move(request), deadline);
+            if (count > _config.maxPushCount()) {
+                return RPCResponse(dto::K23SIStatus::InternalError("request has already been pushed, incumbent txn won in read push"), dto::K23SIReadResponse{});
```
iafuture, line 765: I don't think this is an internal error. Let's just give back an AbortConflict

src/k2/module/k23si/Module.cpp
```diff
@@ -749,13 +754,17 @@ K23SIPartitionModule::handleRead(dto::K23SIReadRequest&& request, FastDeadline d
         return _makeReadOK(rec);
     }
 
+    count++;
     // record is still pending and isn't from same transaction.
     return _doPush(request.key, versions.WI->data.timestamp, request.mtr, deadline)
```
iafuture, line 759: I think this overall logic should be done in one place - inside _doPush itself. I'd rather pass this counter in there  and generate the correct response before attempting actual work.

That way it also works across the board and you don't have to duplicate this code for read/write/query

iafuture (changes requested):

src/k2/module/k23si/Module.cpp
jfunston: Why not a retryable error?

johnfangAFW: The transaction should be aborted after this error and it is usually caused by high conflict in transactions, let us use AbortConflict here.

src/k2/module/k23si/Module.cpp
```diff
@@ -720,7 +720,12 @@ K23SIPartitionModule::handleQuery(dto::K23SIQueryRequest&& request, dto::K23SIQu
 
 seastar::future<std::tuple<Status, dto::K23SIReadResponse>>
 K23SIPartitionModule::handleRead(dto::K23SIReadRequest&& request, FastDeadline deadline) {
```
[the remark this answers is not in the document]
johnfangAFW: sure, will do.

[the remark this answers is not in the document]
johnfangAFW: removed _handleRead() and use one signature.

src/k2/module/k23si/Module.cpp
```diff
@@ -1112,16 +1121,16 @@ K23SIPartitionModule::handleWrite(dto::K23SIWriteRequest&& request, FastDeadline
                 }
 
                 K2LOG_D(log::skvsvr, "succeeded creating TR. Processing write for {}", request.mtr);
-                return _processWrite(std::move(request), deadline);
+                return _processWrite(std::move(request), deadline, 0);
```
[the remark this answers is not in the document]
johnfangAFW: this is needed and we need to pass the initial count as zero..

src/k2/module/k23si/Module.cpp
johnfangAFW: changed to AbortConflict

src/k2/module/k23si/Module.cpp
johnfangAFW: good suggestion, moved the check logic to _doPush().

iafuture (approved):

jfunston (approved):

PR #259, opened 2021-12-14
CPO select nodes for collection assignment

jfunston:
CPO will select nodepool nodes for collection assignment (drawn from "nodepool_endpoints" cli arg), rather than clients specifying endpoints. For hash collections the number of nodes comes from the "minNodes" field of the collection metadata. For range collections the number of nodes is equal to the number of range ends passed in.

The nodepool process now gets the CPO endpoint from the partition assignment request instead of from the cli arg.

The TSO client will now also get the TSO server endpoint from the CPO. The CPO gets the endpoints from the "tso_endpoints" cli arg.

The nodepool still gets its persistence endpoints directly from its command line, because we need a way to distribute the load among persistence servers, which is currently handled by the cluster running scripts using the command line options.

Tested with unit tests and integration tests. Repos using the platform may need to be updated since some command line arguments have been changed or removed, and the interface for creating collections from the k23si client has changed.

### DIFF
--- a/src/k2/assignmentManager/AssignmentManager.cpp
+++ b/src/k2/assignmentManager/AssignmentManager.cpp
@@ -64,8 +64,6 @@ seastar::future<> AssignmentManager::start() {
 seastar::future<std::tuple<Status, dto::AssignmentCreateResponse>>
 AssignmentManager::handleAssign(dto::AssignmentCreateRequest&& request) {
     K2LOG_I(log::amgr, "Received request to create assignment in collection {}, for partition {}", request.collectionMeta.name, request.partition);
-    // TODO, consider current load on all cores and potentially re-route the assignment to a different core
-    // for now, simply pass it onto local handler
 
     dto::CollectionMetadata& meta = request.collectionMeta;
     dto::Partition& partition = request.partition;
@@ -99,7 +97,7 @@ AssignmentManager::handleAssign(dto::AssignmentCreateRequest&& request) {
     }
 
     _collectionName = meta.name;
-    _pmodule = std::make_unique<K23SIPartitionModule>(std::move(meta), partition);
+    _pmodule = std::make_unique<K23SIPartitionModule>(std::move(meta), partition, request.cpoEndpoint);
     return _pmodule->start().then([partition = std::move(partition)] () mutable {
         if (partition.endpoints.size() > 0) {
             partition.astate = dto::AssignmentState::Assigned;
--- a/src/k2/assignmentManager/AssignmentManager.cpp
+++ b/src/k2/assignmentManager/AssignmentManager.cpp
@@ -29,6 +29,7 @@ Copyright(c) 2020 Futurewei Cloud
 #include <k2/dto/MessageVerbs.h>         // our DTO
 #include <k2/transport/RPCDispatcher.h>  // for RPC
 #include <k2/transport/Status.h>         // for RPC
+#include <k2/transport/Discovery.h>      // for selectBestEndpointString
 
 namespace k2 {
 
@@ -96,15 +97,7 @@ AssignmentManager::handleAssign(dto::AssignmentCreateRequest&& request) {
         partition.endpoints.insert(rdma_ep->url);
     }
 
-    String cpoEP = "";
-    for (const String& ep : request.cpoEndpoints) {
-        if (ep.find(RRDMARPCProtocol::proto) != String::npos && rdma_ep) {
-            cpoEP = ep;
-        }
-        else if (ep.find(TCPRPCProtocol::proto) != String::npos && !rdma_ep) {
-            cpoEP = ep;
-        }
-    }
+    String cpoEP = Discovery::selectBestEndpointString(request.cpoEndpoints);
     K2LOG_I(log::amgr, "From CPO: {} chose {}", request.cpoEndpoints, cpoEP);
 
     _collectionName = meta.name;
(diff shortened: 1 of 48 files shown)


src/k2/assignmentManager/AssignmentManager.cpp
```diff
@@ -98,8 +96,19 @@ AssignmentManager::handleAssign(dto::AssignmentCreateRequest&& request) {
         partition.endpoints.insert(rdma_ep->url);
     }
 
+    String cpoEP = "";
```
iafuture, line 99: The CPO endpoint should come as an auto-rrdma, not tcp/rdma directly. Otherwise, we won't be able to restart the CPO.

src/k2/cmd/httpclient/http_client.cpp
```diff
@@ -162,8 +162,21 @@ class HTTPClient {
         if (myid == 0) {
             K2LOG_I(k2::log::httpclient, "Creating collection...");
             _startFut = _startFut.then([this] {
-                return _client.makeCollection(collname).discard_result()
-                .then([this] () {
+                k2::dto::CollectionMetadata meta {
```
iafuture, line 165: I think I missed this before - what is this collection we're creating in the HTTP client?

src/k2/cpo/service/HealthMonitor.h
```diff
@@ -102,6 +103,10 @@ class HealthMonitor {
     // required for seastar::distributed interface
     seastar::future<> gracefulStop();
     seastar::future<> start();
+
+    std::vector<String> getNodepoolEndpoints();
```
iafuture, line 107: return const& instead?

src/k2/cpo/service/Service.cpp
```diff
@@ -132,21 +152,49 @@ seastar::future<> CPOService::start() {
...
 
-    if (rangeEnds.size() != eps.size()) {
-        K2LOG_W(log::cposvr, "Error in client's make collection request: collection endpoints size does not equal rangeEnds size");
-        return -1;
+        return seastar::make_ready_future<>();
+    });
+}
+
+String CPOService::_assignToFreeNode(String collection) {
+    auto it = _nodesToCollection.begin();
```
iafuture, line 169: I think we should be writing _nodeToCollection to disk somewhere here, and also load it from disk on start.  I wonder if we should also start thinking about our persistence story in the CPO, at least add the initial interface

src/k2/cpo/service/Service.cpp
```diff
@@ -420,6 +486,14 @@ void CPOService::_assignCollection(dto::Collection& collection) {
         dto::AssignmentCreateRequest request;
         request.collectionMeta = collection.metadata;
         request.partition = part;
+        auto tcp_ep = k2::RPC().getServerEndpoint(k2::TCPRPCProtocol::proto);
```
iafuture, line 489: auto-rrdma?

src/k2/cpo/service/Service.h
```diff
@@ -34,10 +34,18 @@ Copyright(c) 2020 Futurewei Cloud
...
 namespace k2::cpo {
-namespace log {
-inline thread_local k2::logging::Logger cposvr("k2::cpo_service");
-}
+
+// Used by the CPOService class to track mapping of nodes to assigned collections
+class NodeAssignmentEntry {
+public:
+    String collection;
+    bool assigned;
```
iafuture, line 46: add default for the bool: {false}

src/k2/module/k23si/Module.cpp
```diff
@@ -371,15 +373,20 @@ seastar::future<> K23SIPartitionModule::start() {
...
+    _cpo.init(_cpoEndpoint);
     if (_cmeta.retentionPeriod < _config.minimumRetentionPeriod()) {
         K2LOG_W(log::skvsvr,
             "Requested retention({}) is lower than minimum({}). Extending retention to minimum",
             _cmeta.retentionPeriod, _config.minimumRetentionPeriod());
         _cmeta.retentionPeriod = _config.minimumRetentionPeriod();
     }
 
-    return _tsoClient.getTimestamp()
+    return AppBase().getDist<tso::TSOClient>().invoke_on(seastar::this_shard_id(), &tso::TSOClient::bootstrap,
```
iafuture, line 385: _tsoClient is already the local instance - I think you can just call _tsoClient.bootstrap(_cpoEndpoint) here.

src/k2/tso/client/Client.cpp
```diff
@@ -48,14 +49,41 @@ void TSOClient::_registerMetrics() {
 }
 
 seastar::future<> TSOClient::start() {
```
iafuture, line 51: I guess we can leave it for now, but I don't think we need the dist-based TSO client anymore so this init sequence can go away.

iafuture (changes requested):

src/k2/assignmentManager/AssignmentManager.cpp
jfunston: Add transport helper for ep string selection

src/k2/cpo/service/Service.cpp
jfunston: use getServerEndpoints

src/k2/cpo/service/HealthMonitor.h
jfunston: Use invoke_on internally

src/k2/cpo/service/Service.cpp
jfunston: Add TODO comment for CPO persistence with "fileutil::writeFile"

src/k2/cpo/service/HealthMonitor.h
jfunston: invoke_on requires a member function so I couldn't use a lambda, but the helper functions are private

iafuture (approved):

Issue #270, opened 2022-03-18
Improve testing for HTTP proxy client


jfunston:
The HTTP proxy client receives requests from HTTP in JSON format and forwards the requests through the native C++ chogori client. Currently it only supports a static collection and schemas with begin txn, read, write, and end txn support. First we should improve the testing of the existing functionality.

- Python test script: test/integration/test_http.py
- Runner test script: test/integration/test_http_client.sh
- HTTP proxy client implementation: src/k2/cmd/httpclient/http_client.cpp
- C++ native client implementation: src/k2/module/k23si/client/k23si_client.h

Tests to add:
- Read/Write/End with a bad txn id.
- Read/write with a bad schemaName
- Write with a bad schemaVersion
- Read a non-existent key
- Using two txns, a write-after-read conflict
- Read/write with a field specified that is not part of the schema
- Read/write with a field where the value type does not match the schema

ahsank:
@jfunston The issue is fixed.

PR #278, opened 2022-03-30
Incorporate seastar fix to not enclose json response in array

ahsank:
Related fix: https://github.com/futurewei-cloud/chogori-seastar-rd/pull/14. Also reduce number of times string copied. Add more unit tests.
https://github.com/futurewei-cloud/chogori-platform/issues/270


test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
...
+            record = {partitionKeyName: partition, rangeKeyName: range_key,
+                      dataFieldName: data}
+            request = {"collectionName": coll, "schemaName": schema, "txnID": txn, "schemaVersion": ver, "record": record}
+        elif api == "/api/Read":
+            record = {partitionKeyName: partition, rangeKeyName: range_key}
+            request = {"collectionName": coll, "schemaName": schema, "txnID": txn, "schemaVersion": ver, "record": record}
+        elif api == "/api/EndTxn":
+            # Commit
+            request = {"txnID": txn, "commit": True}
+            print(request)
```
jfunston, line 93: leftover print?

test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
...
+        url = args.http + api
+        r = requests.post(url, data=json.dumps(request))
+        result = r.json()
+        print(result)
+        return result
+
+    def test_schema_validation(self):
+        # Begin Txn
+        data = {}
+        url = args.http + "/api/BeginTxn"
```
jfunston, line 106: You can use your sendReq helper for the begin too

test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
...
+        r = requests.post(url, data=json.dumps(data))
+        result = r.json()
+        result = result[0][0]
+        print(result)
+        self.assertEqual(result["status"]["code"], 201);
+        ID = result["txnID"]
+
+        # Write with bad Txn ID, should fail
+        result = self.sendReq("/api/Write", ID+1)
+        result = result[0]
```
jfunston, line 116: Can you look into the http proxy server code and see why the response is sometimes double nested and sometimes not? If we can fix that for this PR that would be great.

test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
...
+        result = self.sendReq("/api/Write", ID)
+        result = result[0][0]
+        self.assertEqual(result["status"]["code"], 201)
+
+        # Read with bad Txn ID, should fail
+        result = self.sendReq("/api/Read", ID+1)
+        result = result[0]
+        self.assertEqual(result["status"]["code"], 400)
+
+        # Read with bad collectionName, should fail
```
jfunston, line 161: Why are some of the tests repeated here?

test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
...
+        result = result[0][0]
+        self.assertEqual(result["status"]["code"], 200)
+
+        # Commit again, should fail
+        result = self.sendReq("/api/EndTxn", ID)
+        result = result[0]
+        self.assertEqual(result["status"]["code"], 400)
+
+    # Test read write conflict between two transactions
+    def test_read_write_txn(self):
```
jfunston, line 200: To get this test to work you need a sequence like:

- Begin Txn1
- Begin Txn2
- Read key with Txn2 (404 expected)
- Write key with Txn1 (conflict error code expected)

What you have is trying to read a record with an old timestamp, which is not an error for us because of MVCC. I can explain more if you want.

jfunston (changes requested):
For the test of having a schema field name that is not part of the schema, you will need to handle that case in the http proxy server code, but you can do it as a separate PR if you want.

test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
...
+        else:
+            raise ValueError('Invalid api' + api)
+
+        url = args.http + api
+        r = requests.post(url, data=json.dumps(request))
+        result = r.json()
+        print(result)
+        return result
+
+    def test_schema_validation(self):
```
iafuture, line 103: not sure why this is a test called "schema_validation". It is doing transaction ops, not validating schema

test/integration/test_http.py
```diff
@@ -73,5 +73,196 @@ def test_basicTxn(self):
         print(result)
         self.assertEqual(result["status"]["code"], 200);
 
+    # Helper method to populate and send request with default values
```
iafuture, line 76: I find the test difficult to read as all it does is making calls to a single function with 11 fields, most of which with default values that aren't very intuitive. There are also a lot of repeated string constants that I can't tell if you meant them to be what they are or you have a copy/paste error

Why not create helpers with API similar to a client library instead? We can actually develop this  as an http client for python.

```python
class QueryResponse:

class Txn:
  def __init__(self, address): pass
  # return (status, record)
  def read(self, record): pass
  # return (status)
  def write(self, record): pass
  # return (status, QueryResponseObj)
  def scan(self, query): pass

class Client:
  # return (status, txnObj)
  def beginTxn(self): pass

  # return (status)
  def createCollection(self, collname, ranges,....): pass

  # return (status, schemaObj)
  def createSchema(self, collname, schemaName, ....): pass
```

then a test would be, e.g.
``` python
def test_conflict(self):
    coll1 = 'testCollection01'
    schema1 = 'testSchema01'
    client = Client(args.http)
    txn1 = client.beginTxn()
    txn2 = client.beginTxn()

    status = txn1.write(coll1, schema1, {'pkey': 'partitionKey1',
               'rkey': 'rangeKey1', 'field1': 'abcd', 'field2': 'efgh'})
    self.assertEqual(status.code, 201)
    status = txn2.write(coll1, schema1, {'pkey': 'partitionKey1',
               'rkey': 'rangeKey1', 'field1': 'abcd', 'field2': 'efgh'})
    self.assertEqual(status.code, 409)

```

iafuture (changes requested):

test/integration/test_http.py
ahsank: The previous one is for write, this one for read

test/integration/test_http.py
ahsank: Fixed. Populated initial value so that step 3 succeeds.

test/integration/test_http.py
ahsank: Changed to `test_validation`

test/integration/test_http.py
ahsank: Working on it.

ahsank: Added initial version of python client in `skvclient.py`. It's mainly to simplify calling read/write api without passing all  parameters as tests call apis with same parameter except one parameter different.

jfunston (approved):

test/integration/skvclient.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @property
+    def txn_id(self) -> int:
+        return self._txn_id
+
+    def _send_req(self, api, request):
+        url = self._client.http + api
+        r = requests.post(url, data=json.dumps(request))
+        result = r.json()
+        print(result)
```
iafuture, line 88: probably don't want to print here?

test/integration/skvclient.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(self, url: str):
+        self.http = url
+
+    def begin_txn(self) -> Tuple[Status, Txn]:
+        data = {}
+        url = self.http + "/api/BeginTxn"
+        r = requests.post(url, data=json.dumps(data))
+        result = r.json()
+        print(result)
```
iafuture, line 126: and probably no print here either

iafuture (changes requested):
I think this looks much better. You should update the PR description for a better commit message that lists the main contributions here

src/k2/infrastructure/APIServer.h
```diff
@@ -49,14 +49,23 @@ class api_route_handler : public seastar::httpd::handler_base  {
         // seastar returns a 200 status code by default, which is what we want. The k2 status code will
         // be embedded in the returned json object
         return _handler(req->content)
-        .then([rep=std::move(rep)] (String&& json) mutable {
-            rep->write_body("json", json);
+        .then([rep=std::move(rep)] (nlohmann::json&& json) mutable {
+            rep->write_body("json",  [json=std::move(json)] (auto&& os)  {
```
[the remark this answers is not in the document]
ahsank: Fixed

ahsank:
Updated commit message.

iafuture (approved):

PR #282, opened 2022-04-08
Create, get schema, and create collection by http proxy

ahsank:
https://github.com/futurewei-cloud/chogori-platform/issues/273

### DIFF
--- a/src/k2/dto/ControlPlaneOracle.h
+++ b/src/k2/dto/ControlPlaneOracle.h
@@ -139,6 +139,18 @@ struct CreateSchemaResponse {
     K2_DEF_FMT(CreateSchemaResponse);
 };
 
+static constexpr int64_t ANY_VERSION = -1;
+
+// Request to get a schema
+struct GetSchemaRequest {
+    String collectionName;
+    String schemaName;
+    int64_t schemaVersion{ANY_VERSION };
+
+    K2_PAYLOAD_FIELDS(collectionName, schemaName, schemaVersion);
+    K2_DEF_FMT(GetSchemaRequest, collectionName, schemaName, schemaVersion);
+};
+
 // Get all versions of all schemas associated with a collection
 struct GetSchemasRequest {
     String collectionName;
--- a/src/k2/dto/ControlPlaneOracle.h
+++ b/src/k2/dto/ControlPlaneOracle.h
@@ -139,13 +139,13 @@ struct CreateSchemaResponse {
     K2_DEF_FMT(CreateSchemaResponse);
 };
 
-static constexpr int64_t ANY_VERSION = -1;
+static inline constexpr int64_t ANY_SCHEMA_VERSION = -1;
 
 // Request to get a schema
 struct GetSchemaRequest {
     String collectionName;
     String schemaName;
-    int64_t schemaVersion{ANY_VERSION };
+    int64_t schemaVersion{ANY_SCHEMA_VERSION};
 
     K2_PAYLOAD_FIELDS(collectionName, schemaName, schemaVersion);
     K2_DEF_FMT(GetSchemaRequest, collectionName, schemaName, schemaVersion);
--- a/src/k2/httpproxy/HTTPProxy.cpp
+++ b/src/k2/httpproxy/HTTPProxy.cpp
@@ -380,7 +380,7 @@ seastar::future<nlohmann::json> HTTPProxy::_handleWrite(nlohmann::json&& request
 
 seastar::future<std::tuple<k2::Status,dto::CreateSchemaResponse>> HTTPProxy::_handleCreateSchema(
     dto::CreateSchemaRequest&& request) {
-    K2LOG_D(k2::log::httpproxy, "Received schema request {}", request);
+    K2LOG_D(k2::log::httpproxy, "Received create schema request {}", request);
     return _client.createSchema(std::move(request.collectionName), std::move(request.schema))
         .then([] (CreateSchemaResult&& result) {
             return seastar::make_ready_future<std::tuple<k2::Status,dto::CreateSchemaResponse>>(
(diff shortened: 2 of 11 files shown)


ahsank:
@jfunston @iafuture The PR is ready for review.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -378,6 +378,31 @@ seastar::future<nlohmann::json> HTTPProxy::_handleWrite(nlohmann::json&& request
     });
 }
 
+seastar::future<std::tuple<k2::Status,dto::CreateSchemaResponse>> HTTPProxy::_handleCreateSchema(
+    dto::CreateSchemaRequest&& request) {
+    K2LOG_D(k2::log::httpproxy, "Received create schema request {}", request);
+    return _client.createSchema(std::move(request.collectionName), std::move(request.schema))
+        .then([] (CreateSchemaResult&& result) {
+            return seastar::make_ready_future<std::tuple<k2::Status,dto::CreateSchemaResponse>>(
```
iafuture, line 386: This is where we use the helper func from the dispatcher. Instead of the seastar::make_future....., just use:
```
return RPCResponse(std::move(result.status), dto::CreateSchemaResponse{});
```

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -378,6 +378,31 @@ seastar::future<nlohmann::json> HTTPProxy::_handleWrite(nlohmann::json&& request
...
+                std::make_pair(result.status, dto::CreateSchemaResponse()));
+        });
+}
+
+seastar::future<std::tuple<k2::Status, dto::Schema>> HTTPProxy::_handleGetSchema(dto::GetSchemaRequest&& request) {
+    K2LOG_D(k2::log::httpproxy, "Received get schema request {}", request);
+    return _client.getSchema(std::move(request.collectionName), std::move(request.schemaName), request.schemaVersion)
+    .then([](k2::GetSchemaResult&& result) mutable {
+        if (!result.schema) {
+            return seastar::make_ready_future<std::tuple<k2::Status, dto::Schema>>(
```
iafuture, line 397: same comment about using RPCResponse here and next line. Make sure to std::move() the status.

src/k2/httpproxy/HTTPProxy.h
```diff
@@ -26,6 +26,18 @@ Copyright(c) 2022 Futurewei Cloud
 #include <k2/module/k23si/client/k23si_client.h>
 
 namespace k2 {
+namespace dto {
+// Request to get a schema
+struct GetSchemaRequest {
+    String collectionName;
+    String schemaName;
+    uint32_t schemaVersion = 0;
```
iafuture, line 34: This field should be a signed int64. From the client library, there is a special value we use to indicate "any version". You should place it here as a constant:
```
static constexpr int64_t ANY_VERSION = -1;
```

iafuture (changes requested):

test/integration/test_http.py
```diff
@@ -194,5 +194,44 @@ def test_read_write_txn(self):
         status = txn2.end()
         self.assertEqual(status.code, 200)
 
+    def test_schema(self):
```
jfunston, line 197: We don't need to be exhaustive in the testing here since the create schema logic is tested in the C++ tests, but we should test at least a couple failure cases here. Maybe a schema without any key fields specified, a schema with the same field name repeated, and trying to get a schema that doesn't exist.

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Done. It makes it much cleaner!

src/k2/httpproxy/HTTPProxy.h
ahsank: Done

test/integration/test_http.py
ahsank: Added more tests.

ahsank:
Also added create collection api

src/k2/dto/ControlPlaneOracle.h
```diff
@@ -139,6 +139,18 @@ struct CreateSchemaResponse {
     K2_DEF_FMT(CreateSchemaResponse);
 };
 
+static constexpr int64_t ANY_VERSION = -1;
```
iafuture, line 142: I think this needs to be also `inline`

iafuture, line 142: Since this is not in a schema-specific namespace/class, you should name it like so: ANY_SCHEMA_VERSION.

src/k2/dto/ControlPlaneOracle.h
```diff
@@ -139,6 +139,18 @@ struct CreateSchemaResponse {
...
+
+// Request to get a schema
+struct GetSchemaRequest {
+    String collectionName;
+    String schemaName;
+    int64_t schemaVersion{ANY_VERSION };
```
iafuture, line 148: got an extra whitespace inthere

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -378,6 +378,32 @@ seastar::future<nlohmann::json> HTTPProxy::_handleWrite(nlohmann::json&& request
     });
 }
 
+seastar::future<std::tuple<k2::Status,dto::CreateSchemaResponse>> HTTPProxy::_handleCreateSchema(
+    dto::CreateSchemaRequest&& request) {
+    K2LOG_D(k2::log::httpproxy, "Received create schema request {}", request);
```
iafuture, line 383: you are already in the k2 namespace. No need for the k2:: specifier in this file

iafuture: Similarly for dto - you have a `using namespace dto` up top. For c++ files, we are ok with "using" clauses so in the file you don't have to write dto::

test/integration/common_defs.sh
```diff
@@ -1,7 +1,7 @@
 set -e
 CPODIR=/tmp/___cpo_integ_test
 rm -rf ${CPODIR}
-EPS=("tcp+k2rpc://0.0.0.0:10000" "tcp+k2rpc://0.0.0.0:10001" "tcp+k2rpc://0.0.0.0:10002")
+EPS=("tcp+k2rpc://0.0.0.0:10000" "tcp+k2rpc://0.0.0.0:10001" "tcp+k2rpc://0.0.0.0:10002" "tcp+k2rpc://0.0.0.0:10003")
```
iafuture, line 4: why do you need a fourth endpoint?

test/integration/test_http.py
```diff
@@ -27,7 +27,8 @@
 import argparse, unittest, sys
 import requests, json
 from urllib.parse import urlparse
-from skvclient import SKVClient, Txn, DBLoc
+from skvclient import *
```
iafuture, line 30: ok for tests i guess, but in general, don't import *

iafuture (changes requested):

test/integration/common_defs.sh
ahsank: Without it, cpo gives no node available error when I try to create a collection as all three nodes are used by existing `HTTPClient` collection created by HTTPProxy

src/k2/dto/ControlPlaneOracle.h
ahsank: Changed. However it was compiling fine without it.

test/integration/common_defs.sh
jfunston: If you are adding the create collection API as part of this PR then I think you should remove the default collection and schemas that the HTTPProxy creates.

src/k2/dto/ControlPlaneOracle.h
ahsank: Fixed

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Removed namespace from new code.

test/integration/common_defs.sh
ahsank: Done. Deleted code to create default collection/schema from HTTPProxy class and moved the code to python test. Changed node number to 2 for default collection.

iafuture (approved):

PR #285, opened 2022-04-15
Implement Query Api by HTTP Proxy

ahsank:
https://github.com/futurewei-cloud/chogori-platform/issues/275

### DIFF
--- a/src/k2/httpproxy/HTTPProxy.cpp
+++ b/src/k2/httpproxy/HTTPProxy.cpp
@@ -364,19 +364,19 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
 
 seastar::future<nlohmann::json> JsonResponse(Status&& status) {
     nlohmann::json resp;
-    resp["status"] = status;
+    resp["status"] = std::move(status);
     return seastar::make_ready_future<nlohmann::json>(std::move(resp));
 }
 
 seastar::future<nlohmann::json> JsonResponse(Status&& status, nlohmann::json&& response) {
     nlohmann::json jsonResponse;
-    jsonResponse["status"] = status;
+    jsonResponse["status"] = std::move(status);
     jsonResponse["response"] = std::move(response);
     return seastar::make_ready_future<nlohmann::json>(std::move(jsonResponse));
 }
 
 // Get FieldToKeyString value for a type
-template <class T> void getEscapedString(SchemaField& field, nlohmann::json& jsonval,  String& out) {
+template <class T> void getEscapedString(const SchemaField& field, const nlohmann::json& jsonval,  String& out) {
     T val;
     (void)field;
     if constexpr  (std::is_same_v<T, std::decimal::decimal64>
@@ -392,7 +392,7 @@ template <class T> void getEscapedString(SchemaField& field, nlohmann::json& jso
 seastar::future<nlohmann::json> HTTPProxy::_handleGetKeyString(nlohmann::json&& request) {
     String output;
     if (!request.contains("fields"))
-        return JsonResponse(Status{400, "Invalid json"});
+        return JsonResponse(Statuses::S400_Bad_Request("Invalid json"));
 
     for (auto& record : request["fields"]) {
         SchemaField field;
@@ -401,7 +401,7 @@ seastar::future<nlohmann::json> HTTPProxy::_handleGetKeyString(nlohmann::json&& 
         K2_DTO_CAST_APPLY_FIELD_VALUE(getEscapedString, field, record["value"], out);
         output += out;
     }
-    return JsonResponse(Status{200, "success"}, nlohmann::json{{"result", output}});
+    return JsonResponse(Statuses::S200_OK(""), nlohmann::json{{"result", output}});
 }
 
 
@@ -413,7 +413,7 @@ seastar::future<nlohmann::json> HTTPProxy::_handleCreateQuery(nlohmann::json&& j
         jsonReq.at("collectionName").get_to(collectionName);
         jsonReq.at("schemaName").get_to(schemaName);
     } catch(...) {
-        return JsonResponse(Status{400, "Bad json for query request"});
+        return JsonResponse(Statuses::S400_Bad_Request("Bad json for query request"));
     }
 
     return _client.createQuery(collectionName, schemaName)
@@ -448,16 +448,19 @@ seastar::future<nlohmann::json> HTTPProxy::_handleQuery(nlohmann::json&& jsonReq
         jsonReq.at("txnID").get_to(txnID);
         jsonReq.at("queryID").get_to(queryID); 
     } catch(...) {
-        return JsonResponse(Status{400, "Bad json for query request"});
+        return JsonResponse(Statuses::S400_Bad_Request("Bad json for query request"));
     }
-    if (_txns.find(txnID) == _txns.end()) {
-        return JsonResponse(Status{400, "Could not find txnID for query request"});
+    auto txnIter = _txns.find(txnID);
+    if (txnIter == _txns.end()) {
+        return JsonResponse(Statuses::S400_Bad_Request("Could not find txnID for query request"));
     }
-    if (_queries.find(queryID) == _queries.end()) {
-        return JsonResponse(Status{400, "Could not find queryID for query request"});
+    auto queryIter = _queries.find(queryID);
+
+    if (queryIter == _queries.end()) {
+        return JsonResponse(Statuses::S400_Bad_Request("Could not find queryID for query request"));
     }
 
-    return _txns[txnID].query(_queries[queryID])
+    return txnIter->second.query(queryIter->second)
     .then([this, queryID](QueryResult&& result) {
         if(!result.status.is2xxOK()) {
             return JsonResponse(std::move(result.status));
--- a/src/k2/httpproxy/HTTPProxy.cpp
+++ b/src/k2/httpproxy/HTTPProxy.cpp
@@ -362,73 +362,115 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
     });
 }
 
-seastar::future<std::tuple<k2::Status, CreateQueryResponse>> HTTPProxy::_handleCreateQuery(CreateQueryRequest&& request) {
-    return _client.createQuery(request.collectionName, request.schemaName)
-    .then([this, req=std::move(request)] (auto&& result) mutable {
+inline seastar::future<nlohmann::json> JsonResponse(Status&& status) {
+    nlohmann::json resp;
+    resp["status"] = status;
+    return seastar::make_ready_future<nlohmann::json>(std::move(resp));
+}
+
+inline seastar::future<nlohmann::json> JsonResponse(Status&& status, nlohmann::json&& response) {
+    nlohmann::json jsonResponse;
+    jsonResponse["status"] = status;
+    jsonResponse["response"] = std::move(response);
+    return seastar::make_ready_future<nlohmann::json>(std::move(jsonResponse));
+}
+
+// Get FieldToKeyString value for a type
+template <class T> void getEscapedString(SchemaField& field, nlohmann::json& jsonval,  String& out) {
+    T val;
+    (void)field;
+    if constexpr  (std::is_same_v<T, std::decimal::decimal64>
+        || std::is_same_v<T, std::decimal::decimal128>) {
+        throw k2::dto::TypeMismatchException("decimal type not supported with JSON interface");
+    } else {
+        jsonval.get_to(val);
+        out = FieldToKeyString<T>(val);
+    }
+}
+
+// Convert [{type: "FieldType", value: "value"}..] to string that can be used in
+// collection range end
+seastar::future<nlohmann::json> HTTPProxy::_handleGetKeyString(nlohmann::json&& request) {
+    String output;
+    if (!request.contains("fields"))
+        return JsonResponse(Status{400, "Invalid json"});
+
+    for (auto& record : request["fields"]) {
+        SchemaField field;
+        record.at("type").get_to(field.type);
+        String out;
+        K2_DTO_CAST_APPLY_FIELD_VALUE(getEscapedString, field, record["value"], out);
+        output += out;
+    }
+    return JsonResponse(Status{200, "success"}, nlohmann::json{{"result", output}});
+}
+
+
+seastar::future<nlohmann::json> HTTPProxy::_handleCreateQuery(nlohmann::json&& jsonReq) {
+    std::string collectionName;
+    std::string schemaName;
+
+    try {
+        jsonReq.at("collectionName").get_to(collectionName);
+        jsonReq.at("schemaName").get_to(schemaName);
+    } catch(...) {
+        return JsonResponse(Status{400, "Bad json for query request"});
+    }
+
+    return _client.createQuery(collectionName, schemaName)
+    .then([this, req=std::move(jsonReq)] (auto&& result) mutable {
         if(!result.status.is2xxOK()) {
-            return RPCResponse(std::move(result.status), CreateQueryResponse{0});
+            return JsonResponse(std::move(result.status));
         }
         K2LOG_D(log::httpproxy, "begin query {}", result);
-        if (!req.startScanRecord.is_null()) {
-            serializeRecordFromJSON(result.query.startScanRecord, std::move(req.startScanRecord));
+        if (req.contains("startScanRecord")) {
+            serializeRecordFromJSON(result.query.startScanRecord, std::move(req.at("startScanRecord")));
         }
-        if (!req.endScanRecord.is_null()) {
-            serializeRecordFromJSON(result.query.endScanRecord, std::move(req.endScanRecord));
+        if (req.contains("endScanRecord")) {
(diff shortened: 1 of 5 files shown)


src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,83 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+    }
+    if (_queries.find(request.queryID) == _queries.end()) {
+        return RPCResponse(Status{400, "Could not find queryID for query request"}, QueryResponse());
+    }
+    return _txns[request.txnID].query(_queries[request.queryID])
+    .then([this, req=std::move(request)](QueryResult&& result) {
+        if(!result.status.is2xxOK()) {
+            return RPCResponse(std::move(result.status),  QueryResponse());
+        }
+        QueryResponse response;
```
iafuture, line 424: Similar to json handling in createQuery - the conversion SKVRecord <-> Json, should happen in the helpers to/from_json(), not outside. So the QueryResponse would have a vector<SKVRecords> not jsons.

iafuture, line 424: Similar to json handling in createQuery - the conversion SKVRecord <-> Json, should happen in the helpers to/from_json(), not outside. So the QueryResponse would have a vector<SKVRecords> not jsons.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,83 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+        if (!req.endScanRecord.is_null()) {
+            serializeRecordFromJSON(result.query.endScanRecord, std::move(req.endScanRecord));
+        }
+        if (req.limit != 0) {
+            result.query.setLimit(req.limit);
+        }
+        if (req.reverse) {
+            result.query.setReverseDirection(req.reverse);
+        }
+        _queries[_queryID++] = std::move(result.query);
```
iafuture, line 384: We should have timeouts associated with these entries and  we should stop tracking them if timeout is reached. Same for transactions i think.

iafuture, line 384: We should have timeouts associated with these entries and  we should stop tracking them if timeout is reached. Same for transactions i think.

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,83 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+void to_json(nlohmann::json& j, const QueryResponse& q) {
+    j = nlohmann::json{{"records", q.records} };
+    j["done"] = q.done;
+}
+
+void from_json(const nlohmann::json& j, CreateQueryRequest& q) {
+    j.at("collectionName").get_to(q.collectionName);
+    j.at("schemaName").get_to(q.schemaName);
+
+    if (j.contains("startScanRecord")) {
```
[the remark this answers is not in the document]
ahsank: Changed to raw api.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Changed to raw api.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Created issue https://github.com/futurewei-cloud/chogori-platform/issues/286 to handle it.

ahsank: Created issue https://github.com/futurewei-cloud/chogori-platform/issues/286 to handle it

src/k2/httpproxy/HTTPProxy.cpp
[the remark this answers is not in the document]
ahsank: Changed to raw api

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Changed to raw api

ahsank:
@iafuture @jfunston I have completed the change. Please review again.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,125 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
     });
 }
 
+inline seastar::future<nlohmann::json> JsonResponse(Status&& status) {
```
jfunston, line 365: Not a big deal, but I don't think inline does anything useful here or in the following function

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,125 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+    jsonResponse["status"] = status;
+    jsonResponse["response"] = std::move(response);
+    return seastar::make_ready_future<nlohmann::json>(std::move(jsonResponse));
+}
+
+// Get FieldToKeyString value for a type
+template <class T> void getEscapedString(SchemaField& field, nlohmann::json& jsonval,  String& out) {
+    T val;
+    (void)field;
+    if constexpr  (std::is_same_v<T, std::decimal::decimal64>
```
jfunston, line 382: I think you can remove this condition, anything that FieldToKeyString supports we should support here.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,125 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+        return JsonResponse(Status{400, "Invalid json"});
+
+    for (auto& record : request["fields"]) {
+        SchemaField field;
+        record.at("type").get_to(field.type);
+        String out;
+        K2_DTO_CAST_APPLY_FIELD_VALUE(getEscapedString, field, record["value"], out);
+        output += out;
+    }
+    return JsonResponse(Status{200, "success"}, nlohmann::json{{"result", output}});
```
jfunston, line 405: The key string can contain null and non printable bytes. We should check if the json library is handling that properly or if we need to encode it manually.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,125 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+        }
+        K2LOG_D(log::httpproxy, "begin query {}", result);
+        if (req.contains("startScanRecord")) {
+            serializeRecordFromJSON(result.query.startScanRecord, std::move(req.at("startScanRecord")));
+        }
+        if (req.contains("endScanRecord")) {
+            serializeRecordFromJSON(result.query.endScanRecord, std::move(req.at("endScanRecord")));
+        }
+        if (req.contains("limit")) {
+            result.query.setLimit(req["limit"]);
```
jfunston, line 433: Is it handled properly if this or the reverse fields are missing/invalid types in the json?

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,125 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+    if (_queries.find(queryID) == _queries.end()) {
+        return JsonResponse(Status{400, "Could not find queryID for query request"});
+    }
+
+    return _txns[txnID].query(_queries[queryID])
+    .then([this, queryID](QueryResult&& result) {
+        if(!result.status.is2xxOK()) {
+            return JsonResponse(std::move(result.status));
+        }
+        std::vector<nlohmann::json> records;
```
jfunston, line 466: Can use reserve on this vector since we know the size before hand

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,125 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+            records.push_back(serializeJSONFromRecord(record));
+        }
+        nlohmann::json resp;
+        resp["records"] = std::move(records);
+        resp["done"] = _queries[queryID].isDone();
+        return JsonResponse(std::move(result.status), std::move(resp));
+    });
+}
+
+seastar::future<std::tuple<k2::Status, EmptyResponse>> HTTPProxy::_handleCloseQuery(CloseQueryRequest&& request) {
```
jfunston, line 477: I think we can remove this and automatically close it when we see that it is done or an error in the query request. We can save the close by timer for later.

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Removed inline. Not sure if inline or not inline better.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: It fails even before calling `FieldToKeyString`, when it tries to get value into val variable as  `nlohmann::basic_json<>::get_to(std::decimal::decimal64&)` not defined. I think there is an issue for decimal support.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: It only executes the line if the value is not missing. Also test code doesn't send the value if not specified so that code is tested. Added test to verify invalid type.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Yes. Done.

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp
jfunston: That's good. What about the return path though? The intent is that the user can use this to find the range ends for the collection. Is the string decoded back to the format we expect for the collection creation api? You can make an issue for this if you want to save it for later.

jfunston (dismissed):

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,124 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
     });
 }
 
+seastar::future<nlohmann::json> JsonResponse(Status&& status) {
+    nlohmann::json resp;
+    resp["status"] = status;
```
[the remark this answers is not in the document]

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,124 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+
+seastar::future<nlohmann::json> JsonResponse(Status&& status, nlohmann::json&& response) {
+    nlohmann::json jsonResponse;
+    jsonResponse["status"] = status;
+    jsonResponse["response"] = std::move(response);
+    return seastar::make_ready_future<nlohmann::json>(std::move(jsonResponse));
+}
+
+// Get FieldToKeyString value for a type
+template <class T> void getEscapedString(SchemaField& field, nlohmann::json& jsonval,  String& out) {
```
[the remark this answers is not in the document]
ahsank: Yes. Done

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,124 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+        return JsonResponse(Status{400, "Invalid json"});
+
+    for (auto& record : request["fields"]) {
+        SchemaField field;
+        record.at("type").get_to(field.type);
+        String out;
+        K2_DTO_CAST_APPLY_FIELD_VALUE(getEscapedString, field, record["value"], out);
+        output += out;
+    }
+    return JsonResponse(Status{200, "success"}, nlohmann::json{{"result", output}});
```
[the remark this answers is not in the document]
ahsank: Done. Statuses::S200_OK() doesn't compile used Statuses::S200_OK("")

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,124 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+    try {
+        jsonReq.at("txnID").get_to(txnID);
+        jsonReq.at("queryID").get_to(queryID); 
+    } catch(...) {
+        return JsonResponse(Status{400, "Bad json for query request"});
+    }
+    if (_txns.find(txnID) == _txns.end()) {
+        return JsonResponse(Status{400, "Could not find txnID for query request"});
+    }
+    if (_queries.find(queryID) == _queries.end()) {
```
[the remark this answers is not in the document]
ahsank: Used outside of if to avoid nested condition.

test/integration/common_defs.sh
```diff
@@ -1,7 +1,7 @@
 set -e
 CPODIR=/tmp/___cpo_integ_test
 rm -rf ${CPODIR}
-EPS=("tcp+k2rpc://0.0.0.0:10000" "tcp+k2rpc://0.0.0.0:10001" "tcp+k2rpc://0.0.0.0:10002")
```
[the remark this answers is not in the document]
ahsank: I need new nodes to create collection/schema with range partition to test query api. Current nodes are used by existing tests that uses hash partition.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -362,6 +362,124 @@ seastar::future<std::tuple<Status, CollectionCreateResponse>> HTTPProxy::_handle
...
+        if (req.contains("reverse")) {
+            result.query.setReverseDirection(req["reverse"]);
+        }
+        _queries[_queryID++] = std::move(result.query);
+        nlohmann::json resp{{"queryID", _queryID - 1}};
+        return JsonResponse(std::move(result.status), std::move(resp));
+    });
+}
+
+seastar::future<nlohmann::json> HTTPProxy::_handleQuery(nlohmann::json&& jsonReq) {
```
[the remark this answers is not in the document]
ahsank: Created issue https://github.com/futurewei-cloud/chogori-platform/issues/288

test/integration/common_defs.sh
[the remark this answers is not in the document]
iafuture: I don't think you need this. Other tests run on their own cluster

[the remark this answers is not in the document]
ahsank: I meant other tests in the same python file. They use collection with range partition 
https://github.com/futurewei-cloud/chogori-platform/blob/09605d89fbe5ea398e49d338a6f6f1e66f5b9901/test/integration/test_http.py#L45-L49
One option is I switch using range partition instead of hash partition for all tests. But then we don't be testing that it can create collection with hash partition.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -448,16 +448,19 @@ seastar::future<nlohmann::json> HTTPProxy::_handleQuery(nlohmann::json&& jsonReq
...
     }
-    if (_txns.find(txnID) == _txns.end()) {
-        return JsonResponse(Status{400, "Could not find txnID for query request"});
+    auto txnIter = _txns.find(txnID);
+    if (txnIter == _txns.end()) {
+        return JsonResponse(Statuses::S400_Bad_Request("Could not find txnID for query request"));
     }
-    if (_queries.find(queryID) == _queries.end()) {
-        return JsonResponse(Status{400, "Could not find queryID for query request"});
+    auto queryIter = _queries.find(queryID);
```
iafuture, line 457: you don't need queryIter so why have a variable for it? You can evaluate find() as before in the if condition to make the code a bit easier on the eyes.

test/integration/common_defs.sh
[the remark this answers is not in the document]
iafuture: I see. You should probably use CollectionDrop then. We don't have this now in the proxy, but probably it's something to add next. For now, make an issue to add CollectionDrop support, and as part of that change we can have the tests use it

src/k2/httpproxy/HTTPProxy.cpp
ahsank: I need queryiter  for the query api 6 lines below

https://github.com/futurewei-cloud/chogori-platform/blob/09605d89fbe5ea398e49d338a6f6f1e66f5b9901/src/k2/httpproxy/HTTPProxy.cpp#L463

iafuture: oh yeah, you're right. We should probably combine the two maps when we add support for timeout, but for now, looks good.

test/integration/common_defs.sh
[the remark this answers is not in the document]
ahsank: Added https://github.com/futurewei-cloud/chogori-platform/issues/289 for CollectionDrop support

iafuture (approved):

PR #291, opened 2022-04-20
Support Expression in Query Api in HTTP Proxy

ahsank:


ahsank:
Also changed http proxy behavior to not output NULL for fields absent in records, to avoid removing suchfields by client.

src/k2/dto/Expression.cpp
```diff
@@ -26,6 +26,29 @@ namespace k2 {
 namespace dto {
 namespace expression {
 
+template <class T> void setPayloadValue(Value& target,  const nlohmann::json& jsonval) {
```
jfunston, line 29: I'm still concerned with how we how handle the encoding/decoding of field types. In the existing to_json code, we use our own hex encoding method. I think the way it is now if we convert to json and then back from that json, we will not get the original object back. It probably doesn't matter for this PR but it will need to be fixed in the HTTP C++ client.

jfunston (dismissed):

jfunston (approved):

PR #294, opened 2022-05-04
Cleanup idle transactions after timeout

ahsank:
Also associate queries with transaction in api and data structure.

Followed pattern of https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/transport/RPCDispatcher.h#L268 to implement.

https://github.com/futurewei-cloud/chogori-platform/issues/286

### DIFF
--- a/src/k2/common/MapWithExpiry.h
+++ b/src/k2/common/MapWithExpiry.h
@@ -132,12 +132,14 @@ public:
             });
     }
 
+    typedef typename TimestampOrderedMap<KeyT, ValT>::iterator iterator;
+
     // Check if key is present, with optionally push back expiry if present
-    bool isPresent(const KeyT &key, bool reorder) {
-        auto iter =  _elems.find(key);
-        if (iter == _elems.end()) return false;
+    iterator find(const KeyT &key, bool reorder) {
+        auto iter = _elems.find(key);
+        if (iter == _elems.end()) return iter;
         if (reorder) _elems.resetTs(iter, getExpiry());
-        return true;
+        return iter;
     }
 
     TimePoint getExpiry() {return ClockT::now() + _timeout;}
@@ -145,6 +147,8 @@ public:
     void erase(const KeyT &key) {_elems.erase(key);}
     auto size() {return _elems.size(); }
     ValT& at(const KeyT &key) {return _elems.at(key);}
+    iterator end() {return _elems.end();}
+    ValT& getValue(iterator it) {return it->second.getValue();}
 
 private:
     TimestampOrderedMap<KeyT, ValT> _elems;
@@ -153,5 +157,4 @@ private:
     Duration _timeout = 10s;
 
 };
-
 }
--- a/src/k2/common/MapWithExpiry.h
+++ b/src/k2/common/MapWithExpiry.h
@@ -1,0 +1,157 @@
+/*
+MIT License
+
+Copyright(c) 2022 Futurewei Cloud
+
+    Permission is hereby granted,
+    free of charge, to any person obtaining a copy of this software and associated documentation files(the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and / or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions :
+
+    The above copyright notice and this permission notice shall be included in all copies
+    or
+    substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS",
+    WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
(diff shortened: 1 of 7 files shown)


src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -186,13 +191,41 @@ seastar::future<> HTTPProxy::start() {
...
+
 seastar::future<nlohmann::json> HTTPProxy::_handleBegin(nlohmann::json&& request) {
     (void) request;
     return _client.beginTxn(k2::K2TxnOptions())
     .then([this] (auto&& txn) {
         K2LOG_D(k2::log::httpproxy, "begin txn: {}", txn.mtr());
-        _txns[_txnID++] = std::move(txn);
-        return JsonResponse(Statuses::S201_Created("Begin txn success"), nlohmann::json{{"txnID", _txnID - 1}});
+        auto txnid = _txnID++;
+        seastar::timer<> timer([this, txnid] {
```
jfunston, line 204: I don't think we should have a timer for every txn, they can be a bit expensive in seastar because changing their timeout time is a system call. We should have one timer that tracks the txn that is soonest to timeout. Their is an example of this in the module txn timeouts.

Also you should use the timers in src/k2/common/Timer.h which wraps the seastar timers and fixes some issues.

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -209,12 +218,12 @@ seastar::future<nlohmann::json> HTTPProxy::_handleEnd(nlohmann::json&& request)
...
-    if (it == _txns.end()) {
+    auto iter = _txns.find(id, false);
+    if (iter == _txns.end()) {
         return JsonResponse(Statuses::S400_Bad_Request("Could not find txnID for end request"));
     }
 
-    return it->second.end(commit)
+    return _txns.getValue(iter).txn.end(commit)
     .then([this, id] (k2::EndResult&& result) {
         _txns.erase(id);
```
jfunston, line 228: I think this is a race condition with the timeout timer expiration. You could have two in flight end() calls, which is ok, but the txn will be erased twice. If you make the erase call here safe to call on a non-existing item then it should be fine.

src/k2/common/MapWithExpiry.h
```diff
@@ -0,0 +1,160 @@
...
+                    bgFuts.push_back(iter->second.getValue().cleanup().discard_result());
+                }
+                return seastar::when_all_succeed(bgFuts.begin(), bgFuts.end());
+            });
+    }
+
+    typedef typename TimestampOrderedMap<KeyT, ValT>::iterator iterator;
+
+    // Check if key is present, with optionally push back expiry if present
+    iterator find(const KeyT &key, bool reorder) {
```
jfunston, line 138: Maybe a separate find_and_reorder function instead of the bool? Or an enum instead of the bool? It makes it a bit hard to read the usage of this function as it is.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -209,12 +218,12 @@ seastar::future<nlohmann::json> HTTPProxy::_handleEnd(nlohmann::json&& request)
...
-    std::unordered_map<uint64_t, k2::K2TxnHandle>::iterator it = _txns.find(id);
-    if (it == _txns.end()) {
+    auto iter = _txns.find(id, false);
+    if (iter == _txns.end()) {
         return JsonResponse(Statuses::S400_Bad_Request("Could not find txnID for end request"));
     }
 
-    return it->second.end(commit)
+    return _txns.getValue(iter).txn.end(commit)
     .then([this, id] (k2::EndResult&& result) {
```
jfunston, line 227: The queries need to be cleaned up here too right?

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Good catch! Though the operations seem to run in the same thread, they can interleave? Added more check in the `erase` function. The other place it can happen is `at(...)` function which will throw exception. Leaving it as is, as it's a rare condition and will at worst result in a 500 error caused by uncaught exception instead of 400 error.

src/k2/common/MapWithExpiry.h
ahsank: Fixed.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: When TxnInfo is deallocated, queries will be destroyed as query hash table is inside txn now.

src/k2/httpproxy/HTTPProxy.cpp
jfunston: It is in the same thread but can be concurrent because of the Seastar execution model. The call to end() here ends up as a remote async call. Then the timer could fire and be executed, and after that the .then block here is executed. I think you still need to fix it in the wrapper class because the underlying call to the unordered_map erase with an end iterator is undefined, not an exception.

ahsank: Yes, I added a check inside erase function in the previous update but that got mistakenly reverted. I have put back the check.

Issue #221, opened 2021-09-10
Add pagination of query based on number of rows scanned


jfunston:
This will prevent reactor starvation when a query scans a large number of records but only returns a few to the user due to a filter.

PR #297, opened 2022-05-12
Add config to limit number of rows scanned during query

ahsank:
https://github.com/futurewei-cloud/chogori-platform/issues/221

### DIFF
--- a/src/k2/module/k23si/Module.cpp
+++ b/src/k2/module/k23si/Module.cpp
@@ -457,7 +457,7 @@ Indexer::Iterator K23SIPartitionModule::_initializeScan(const dto::K23SIQueryReq
 
 // Helper for handleQuery. Checks to see if the indexer scan should stop.
 bool K23SIPartitionModule::_isScanDone(const Indexer::Iterator& iter, const dto::K23SIQueryRequest& request,
-                                       size_t response_size) {
+                                       size_t response_size, uint64_t num_scans) {
     // we're at end of iteration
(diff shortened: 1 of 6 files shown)


src/k2/module/k23si/Module.cpp
```diff
@@ -478,21 +478,25 @@ bool K23SIPartitionModule::_isScanDone(const Indexer::Iterator& iter, const dto:
...
+                    const dto::K23SIQueryRequest& request, dto::K23SIQueryResponse& response,
+                    size_t response_size, uint64_t num_scans) {
     auto ikey = iter.getKey();
     // NB: In all checks below, iter.empty signifies that we're past the last
     // key available from the iterator.
     // Three cases where scan is for sure done:
     // 1. Record limit is reached
     // 2. Iterator is not end() but is >= user endKey
     // 3. Iterator is at end() and partition bounds contains endKey
+    (void)num_scans;
```
jfunston, line 499: Can we remove the num_scans parameter from this function instead?

jfunston (changes requested):

src/k2/module/k23si/Module.cpp
ahsank: Yes, I though I will need it. But seems like I don't

ahsank: Fixed

jfunston (approved):

PR #302, opened 2022-05-19
Port k2 unit tests to skv

ahsank:


src/SKVClient/test/SKVRecordTest.cpp
```diff
@@ -0,0 +1,263 @@
...
+    w.write(storage);
+    Binary buf;
+    REQUIRE(w.flush(buf));
+    MPackReader reader(buf);
+    dto::SKVRecord::Storage read_storage{};
+    REQUIRE(reader.read(read_storage));
+    dto::SKVRecord reconstructed("collection", std::make_shared<dto::Schema>(schema),
+                                     std::move(read_storage), true);
+
+    std::optional<String> last = reconstructed.deserializeNext<String>();
```
ahsank, line 255: @iafuture Everything works except this line causes double free error. The stack trace doesn't have symbols. Any suggestion how to debug this issue?

src/SKVClient/src/skv/dto/SKVRecord.cpp
```diff
@@ -266,11 +268,23 @@ SKVRecord SKVRecord::getSKVKeyRecord() {
...
+    }
+
+    // Exclude all value fields
+    for (size_t i = num_keys; i < excludedFields.size(); ++i) {
+        excludedFields[i] = true;
+    }
 
-    for (size_t i = 0; i < schema->partitionKeyFields.size() + schema->rangeKeyFields.size(); ++i) {
-        if (!storage.excludedFields[i]) {
+    for (size_t i = 0; i < excludedFields.size(); ++i) {
```
iafuture, line 286: this should only iterate up to num_keys, not the full number of fields

ahsank: Non key fields needs to have null values.  If it only iterates up to num_keys there won't be any value for non key fields.

iafuture: That's taken care of by the resize call above - it initializes the entire array to false

ahsank: But it's not affecting the builder. The new record is created from builder. One has to populate all fields in builder.

iafuture (approved):

PR #307, opened 2022-06-01
Skv client fixes

iafuture:


src/logging/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required (VERSION 3.13)
+
+project (yule VERSION 0.0.1 LANGUAGES CXX)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+add_compile_options(-O3 -std=c++17 -g -Werror -Wall -Wextra)
```
jfunston, line 6: Not a big deal but I don't think we should set -std here since we are telling cmake to do it above

src/skvhttpclient/CMakeLists.txt
```diff
@@ -1,6 +1,16 @@
 cmake_minimum_required (VERSION 3.13)
 
-project (skvlib VERSION 0.0.1 LANGUAGES CXX)
+project (skvhttpClient VERSION 0.0.1 LANGUAGES CXX)
+
+##
+## MAIN_PROJECT CHECK
+## determine if we are built as a subproject (using add_subdirectory) or as the main project
+##
+set(MAIN_PROJECT OFF)
```
jfunston, line 9: Does this still apply? You mentioned you couldn't get the subproject approach to work.

ahsank (dismissed):

src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
```diff
@@ -111,43 +98,32 @@ SKVRecord::SKVRecord(const String& collection, std::shared_ptr<Schema> s, Storag
...
     partitionKeys.resize(schema->partitionKeyFields.size());
     rangeKeys.resize(schema->rangeKeyFields.size());
 
     uint32_t beginCursor = getFieldCursor();
     seekField(0);
-    for (size_t i = 0; i < partitionKeys.size() + rangeKeys.size() && i < schema->fields.size(); ++i) {
-        // Need the extra 0 argument for vararg macro expansion
-        DO_ON_NEXT_RECORD_FIELD(*this, makeKeyString, 0);
+    auto numKeys = partitionKeys.size() + rangeKeys.size();
+    for (size_t i = 0; i < partitionKeys.size(); ++i) {
```
jfunston, line 108: This refactor assumes that the keys are ordered in the same order as the fields, but that isn't a requirement for us. The only requirement is that all key fields come before all non-key fields.

src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
```diff
@@ -278,8 +242,12 @@ SKVRecord SKVRecord::getSKVKeyRecord() {
     size_t numKeyFields = schema->partitionKeyFields.size() + schema->rangeKeyFields.size();
     for (size_t i = 0; i < schema->fields.size(); ++i) {
         if (i < numKeyFields) {
-            K2_DTO_CAST_APPLY_FIELD_VALUE(_fieldApplier, schema->fields[i], builder, *this);
+            // the partition and range keys are first in the record
+            visitNextField([&builder] (auto&, auto&& value) {
```
jfunston, line 246: Same issue as above, I think this won't work if the key fields are not in order.

jfunston (changes requested):

src/k2/infrastructure/APIServer.h
```diff
@@ -30,31 +30,191 @@ Copyright(c) 2020 Futurewei Cloud
...
+        // see if we can match the incoming content type
+        if (ct == req.ctype) {
+            respCT = ct;
+            break;
+        }
+    }
+
+    if (respCT == ContentType::MSGPACK) {
+        if constexpr(skv::http::isK2SerializableW<Response_T, skv::http::MPackWriter>::value ||
+                     skv::http::isTrivialClass<Response_T>::value) {
```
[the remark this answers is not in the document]
iafuture: I did. I left this here as a compilation error in cases where you want to use a trivial type which isn't K2Serializable. I guess it isn't needed. Will remove

src/logging/CMakeLists.txt
iafuture: Yup. will remove

src/skvhttpclient/CMakeLists.txt
iafuture: It doesn't apply, no. Left over from me trying to get it to work as a subproject. I'll leave it in for now though as it doesn't hurt either and is needed for proper subproject eventually

src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
iafuture: This is a good catch! Fixed

src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
iafuture: I'm not sure this is the same. The order is governed by the schema. The fields however all need to be in the front. This just copies them to the new record, in the same position as they are in the source record. Let's chat to make sure

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -461,44 +414,153 @@ seastar::future<nlohmann::json> HTTPProxy::_handleQuery(nlohmann::json&& jsonReq
...
+    k2::K2TxnOptions opts{
+        .deadline= k2::Deadline<>(request.options.timeout),
+        .priority = static_cast<k2::TxnPriority>(request.options.priority),
+        .syncFinalize = request.options.syncFinalize
+    };
+    return _client.beginTxn(std::move(opts))
+        .then([this](auto&& txn) {
+            K2LOG_D(k2::log::httpproxy, "begin txn: {}", txn.mtr());
+            auto ts = txn.mtr().timestamp;
+            sh::dto::Timestamp shts(ts.tEndTSECount(), ts.tEndTSECount() - ts.tStartTSECount(), ts.tsoId());
```
ahsank, line 495: In   Timestamp ctor https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/dto/Timestamp.h#L47, tsoId is the second parameter
    `Timestamp(uint64_t tEndTSECount, uint32_t tsoId, uint32_t tStartDelta);`
But here it's the third.

src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
jfunston: You are right, this should work.

src/k2/httpproxy/HTTPProxy.cpp
iafuture: good catch! I'll change the Timestamp so it doesn't allow this sort of mistake

iafuture: done

jfunston (approved):

ahsank (approved):

PR #308, opened 2022-06-03
Changed TSO errorbound to be assigned by CPO at startup

ccjeff:
Removed `tso.error_bound` in TSO server and tests. Updated txn wait time in tests (not to wait for too long, otherwise can cause the next txn test's start timestamp fall into the previous test's range).


src/k2/tso/service/Service.cpp
```diff
@@ -143,12 +79,94 @@ seastar::future<> TSOService::_collectWorkerURLs() {
...
 
-seastar::future<> TSOService::_cpoRegister() {
-    // TODO implement CPO-based registration and ID assignment
-    _tsoId = seastar::this_shard_id() + 100000;
+seastar::future<> TSOService::_assign(uint64_t tsoID, k2::Duration errBound) {
+    _CPOErrorBound = errBound;
+    return seastar::make_ready_future<>()
+        .then([this]{
+            seastar::future<> startFut = seastar::make_ready_future<>();
+            if (seastar::this_shard_id() == 0) {
```
jfunston, line 87: I'm wondering if all this shard 0 logic should be part of the assignment logic. If the CPO happens to not assign shard 0 then none of the shards will be able to operate. We can ask Ivan about this too

src/k2/tso/service/Service.cpp
```diff
@@ -143,12 +79,94 @@ seastar::future<> TSOService::_collectWorkerURLs() {
...
+seastar::future<std::tuple<Status, dto::AssignTSOResponse>>
+TSOService::_handleAssignment(dto::AssignTSORequest&& request) {
+    if (request.tsoErrBound <= 0s) {
+        return RPCResponse(Statuses::S400_Bad_Request("TSO error bound cannot be <= 0s"), dto::AssignTSOResponse{});
+    }
+    return _assign(request.tsoID, request.tsoErrBound).then([] {
+        return RPCResponse(Statuses::S200_OK("OK"), dto::AssignTSOResponse{});
+    }).handle_exception([] (auto exp) {
+        K2LOG_W_EXC(log::tsoserver, exp, "failed to assign");
+        return RPCResponse(Statuses::S500_Internal_Server_Error("Unable to assign TSO server"), dto::AssignTSOResponse{});
```
jfunston, line 168: It would be helpful to add the message from the exception here as part of the response

test/integration/test_collection.sh
```diff
@@ -13,10 +13,10 @@ nodepool_child_pid=$!
 persistence_child_pid=$!
 
 # start tso
-./build/src/k2/cmd/tso/tso ${COMMON_ARGS} -c1 --tcp_endpoints ${TSO} --prometheus_port 63003 --tso.error_bound=100us --tso.clock_poller_cpu=${TSO_POLLER_CORE} &
+./build/src/k2/cmd/tso/tso ${COMMON_ARGS} -c1 --tcp_endpoints ${TSO} --prometheus_port 63003 --tso.clock_poller_cpu=${TSO_POLLER_CORE} &
 tso_child_pid=$!
 
-./build/src/k2/cmd/controlPlaneOracle/cpo_main ${COMMON_ARGS} -c1 --tcp_endpoints ${CPO} --data_dir ${CPODIR} --txn_heartbeat_deadline=10s --prometheus_port 63000 --assignment_timeout=1s --nodepool_endpoints ${EPS[@]} --tso_endpoints ${TSO} --persistence_endpoints ${PERSISTENCE} &
+./build/src/k2/cmd/controlPlaneOracle/cpo_main ${COMMON_ARGS} -c1 --tcp_endpoints ${CPO} --data_dir ${CPODIR} --txn_heartbeat_deadline=10s --prometheus_port 63000 --assignment_timeout=1s --nodepool_endpoints ${EPS[@]} --tso_endpoints ${TSO} --tso_error_bound=100us --persistence_endpoints ${PERSISTENCE} &
 cpo_child_pid=$!
```
jfunston, line 21: Maybe we can add a couple error cases to the CPO integration tests. I'm thinking trying to assign a TSO that doesn't exist, and trying to use an error bound that is too low

jfunston (changes requested):

src/k2/cmd/tso/tso_main.cpp
```diff
@@ -31,6 +31,6 @@ int main(int argc, char** argv) {
     app.addApplet<k2::tso::TSOService>();
 
     app.addOptions()("tso.clock_poller_cpu", bpo::value<int16_t>(), "CPU to which to pin the GPS clock polling thread");
-    app.addOptions()("tso.error_bound", bpo::value<k2::ParseableDuration>(), "the error bound for this instance");
+    app.addOptions()("tso.error_bound", bpo::value<k2::ParseableDuration>(), "the error bound for this instance");   // TODO: cleanup?
```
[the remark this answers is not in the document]
ccjeff: Sure!

src/k2/cpo/service/Service.cpp
```diff
@@ -147,7 +139,71 @@ seastar::future<> CPOService::start() {
         }
     }
 
-    return seastar::make_ready_future<>();
+    _tsoAssignTimer.setCallback([this] {
+        return _assignAllTSOs();
```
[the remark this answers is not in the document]
ccjeff: I'm thinking the TSO assignment should be independent from the CPO startup and the CPO should be able to startup and do RPC registration without any TSO.

src/k2/cpo/service/Service.cpp
```diff
@@ -147,7 +139,71 @@ seastar::future<> CPOService::start() {
...
+    });
+    _tsoAssignTimer.arm(0s);
+    
+    RPC().registerRPCObserver<dto::GetTSOEndpointsRequest, dto::GetTSOEndpointsResponse>(dto::Verbs::CPO_GET_TSO_ENDPOINTS, [this] (dto::GetTSOEndpointsRequest&& ) {
+        K2LOG_I(log::cposvr, "healthy TSOs: {}", _healthyTSOs);
+        if (_healthyTSOs.size() > 0) {
+            return RPCResponse(Statuses::S200_OK("GetTSOEndpoints success"),dto::GetTSOEndpointsResponse{_healthyTSOs});
+        } else {
+            return RPCResponse(Statuses::S500_Internal_Server_Error("Unable to get TSO endpoints: no healthy TSO"), dto::GetTSOEndpointsResponse());
```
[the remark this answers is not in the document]
ccjeff: I think maybe 503 is the more appropriate error code here. I think 200 with no endpoints implies response is OK and can be cached whereas we don't want empty endpoints to be cached here.

src/k2/tso/service/Service.cpp
ccjeff: Yeah it make sense. Ivan talked with me about this and I think I'll update this to use a `test_and_set` condition on an atomic flag.

src/k2/cpo/service/Service.cpp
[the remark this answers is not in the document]
jfunston: That makes sense but you don't have to put it in a timer to do that. You can make the call and save the future somewhere, the startup logic doesn't need to wait for it to finish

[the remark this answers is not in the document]
ccjeff: I see. I'm thinking maybe the timer can be used for retry purpose (like keep retrying with some strategy when some TSO endpoints failed to be assigned) but I haven't come up with a clear plan for doing this retry with the timer here. I have changed it to use a future but maybe in the next PR I'll add the retry logic. What do you think?

src/k2/cpo/service/Service.cpp
```diff
@@ -173,39 +173,47 @@ seastar::future<> CPOService::_assignAllTSOs() {
...
-            K2LOG_I(log::cposvr, "cpo retrying assignment for endpoint: {} due to {}", ep, status);
-            return seastar::sleep(_assignTimeout())
-            .then([&ep, tsoID, retry, this] {
-                return _assignTSO(ep, tsoID, retry-1);
-            });
-        }
-        else {
-            K2LOG_W(log::cposvr, "tso assignment unsucessful for endpoint: {} due to {}", ep, status);
+seastar::future<> CPOService::_doAssignTSO(String &ep, size_t tsoID) {
+    return seastar::do_with(k2::dto::AssignTSORequest{.tsoID=tsoID+1000, .tsoErrBound=_TSOErrorBound()}, [&ep, tsoID, this](auto &request) {
```
jfunston, line 177: callRPC does not require the request to live beyond the calling context (it copies data to a payload), so this do_with clause is not needed. Removing it helps readability a bit

jfunston (changes requested):

test/integration/cpo_tso_assign.py
```diff
@@ -0,0 +1,81 @@
...
+        time.sleep(1)
+        new_tso_args = shlex.split(args.cmd)
+        p = subprocess.Popen(new_tso_args)
+        time.sleep(5)
+
+        scount, fcount = readMetrics()
+        self.assertEqual(scount, 3)
+        self.assertEqual(fcount, 0)
+
+        p.kill()
```
iafuture, line 78: should also do p.wait after kill to make sure it exits, perhaps even check to make sure exit code is 0

iafuture (changes requested):

test/k23si/3SITxnTest.cpp
```diff
@@ -1910,14 +1912,17 @@ seastar::future<> testScenario05() {
                         auto& [status, resp] = response;
                         K2EXPECT(log::k23si, status, dto::K23SIStatus::OK);
                     });
```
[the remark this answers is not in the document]
ccjeff: Sure!

test/integration/cpo_tso_assign.py
ccjeff: I have added the p.wait for the subprocess. However it seems like the exit code from the python subprocess module does not match the POSIX exit codes. I can only test if the process is terminated by the signal I sent. https://docs.python.org/2/library/subprocess.html#subprocess.Popen.returncode

iafuture: not sure i understand. i'd do:
`
rcode = p.wait()
self.assertEqual(rcode, 0)
`

ccjeff: Yes this is exactly what i thought it should return but the `rcode` would be -9 instead of 0 since p.wait() just set and returns the `p.returncode` variable according to https://docs.python.org/2/library/subprocess.html#subprocess.Popen.wait. The variable is set to -9, which means it was stopped by a SIGKILL.

iafuture: oh i see. The problem is that you're using SIGKILL. You need to send SIGTERM for graceful shutdown

iafuture: I was using the posix speak for kill() which just means deliver a signal. It seems the Popen API has a separate send_signal() method to use instead of kill()

ccjeff: Ahh that make sense. Thanks!

iafuture (approved):
LGTM! Please wait for Justin to also give feedback

src/k2/cpo/service/Service.cpp
```diff
@@ -147,9 +163,86 @@ seastar::future<> CPOService::start() {
...
+    k2::dto::AssignTSORequest request{.tsoID = tsoID, .tsoErrBound=_TSOErrorBound()};
+    auto txep = RPC().getTXEndpoint(ep);
+    K2LOG_D(log::cposvr, "assigning TSO: {}", ep);
+    if (!txep) {
+        K2LOG_W(log::cposvr, "unable to obtain endpoint for {}", ep);
+        return seastar::make_exception_future(StopRetryException{});
+    }
+    return RPC().callRPC<dto::AssignTSORequest, dto::AssignTSOResponse>(
+        dto::Verbs::TSO_ASSIGNMENT, request, *txep, _assignTimeout()
+    ).then([&ep,tsoID,this] (auto &&result) {
```
jfunston, line 209: I don't think it is safe to capture ep by reference here, you can capture by value instead.

jfunston (changes requested):

jfunston (approved):

iafuture (approved):

PR #309, opened 2022-06-06
implement data apis. add schema cache

iafuture:
This change implements the read/write paths in the python tester to make sure the HTTP proxy APIs are tested

### DIFF
--- a/src/k2/httpproxy/HTTPProxy.cpp
+++ b/src/k2/httpproxy/HTTPProxy.cpp
@@ -395,9 +395,17 @@ HTTPProxy::HTTPProxy() : _client(K23SIClientConfig()) {
 }
 
 seastar::future<> HTTPProxy::gracefulStop() {
-    _txns.clear();
-
-    return seastar::make_ready_future();
+    std::vector<seastar::future<>> futs;
+    for (auto& [ts, txn]: _txns) {
+        futs.push_back(txn.handle.kill());
+    }
+    return seastar::when_all_succeed(futs.begin(), futs.end()).discard_result()
+    .then([this] {
+        _txns.clear();
+    })
+    .then([this] {
+        return _client.gracefulStop();
+    });
 }
 
 seastar::future<> HTTPProxy::start() {
--- a/src/k2/httpproxy/HTTPProxy.cpp
+++ b/src/k2/httpproxy/HTTPProxy.cpp
@@ -32,179 +32,51 @@ inline thread_local k2::logging::Logger httpproxy("k2::httpproxy");
 }
 
 namespace k2 {
-using namespace dto;
-/*
-template <typename T>
-void serializeFieldFromJSON(const k2::SchemaField& field, k2::SKVRecord& record,
-                                   const nlohmann::json& jsonRecord) {
-    T value;
-    jsonRecord.at(field.name).get_to(value);
-    record.serializeNext<T>(value);
-}
 
-template <>
-void serializeFieldFromJSON<k2::String>(const k2::SchemaField& field,
-                                            k2::SKVRecord& record, const nlohmann::json& jsonRecord) {
-    std::string value;
-    jsonRecord.at(field.name).get_to(value);
-    record.serializeNext<k2::String>(value);
-}
-
-template <>
-void serializeFieldFromJSON<std::decimal::decimal64>(const k2::SchemaField& field,
-                                            k2::SKVRecord& record, const nlohmann::json& jsonRecord) {
-    (void) field;
-    (void) record;
-    (void) jsonRecord;
-    throw k2::dto::TypeMismatchException("decimal64 type not supported with JSON interface");
-}
-
-template <>
-void serializeFieldFromJSON<std::decimal::decimal128>(const k2::SchemaField& field,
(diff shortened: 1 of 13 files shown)


src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -476,92 +245,215 @@ HTTPProxy::_handleCreateSchema(sh::dto::CreateSchemaRequest&& request) {
...
+            }
+            return seastar::make_ready_future<std::tuple<Status, std::shared_ptr<dto::Schema>, std::shared_ptr<shd::Schema>>>(std::move(result.status), std::move(result.schema), std::move(shdSchemaPtr));
+        });
+}
+
+HTTPProxy::HTTPProxy() : _client(K23SIClientConfig()) {
+}
+
+seastar::future<> HTTPProxy::gracefulStop() {
+    _txns.clear();
```
ahsank, line 398: Should it abort the txns like before?

iafuture: This functionality should be implemented by the k23si client itself - when it shuts down, it should cleanup any in-progress txns. There is no reason why the user of the client library has to worry about it.

ahsank (approved):

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -476,92 +245,223 @@ HTTPProxy::_handleCreateSchema(sh::dto::CreateSchemaRequest&& request) {
...
+            auto& [status, k2Schema, shdSchema] = schemas;
+            if (!status.is2xxOK()) {
+                return MakeHTTPResponse<shd::WriteResponse>(sh::Status{.code=status.code, .message=status.message}, shd::WriteResponse{});
+            }
+            auto it = _txns.find(request.timestamp);
+            if (it == _txns.end()) {
+                return MakeHTTPResponse<shd::WriteResponse>(sh::Statuses::S410_Gone("transaction does not exist"), shd::WriteResponse{});
+            }
+            dto::SKVRecord k2record(request.collectionName, k2Schema);
+            bool haveKeys = !request.isDelete && request.fieldsForPartialUpdate.size() == 0;
```
jfunston, line 291: We need the keys even if it is a delete or partial update in order to route. The K2 client will also verify the keys are present in the k2record as part of the write, so we can probably just pass true to the constructor.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -476,92 +245,223 @@ HTTPProxy::_handleCreateSchema(sh::dto::CreateSchemaRequest&& request) {
...
+                }
+                auto it = _txns.find(request.timestamp);
+                if (it == _txns.end()) {
+                    return MakeHTTPResponse<shd::ReadResponse>(sh::Statuses::S410_Gone("transaction does not exist"), shd::ReadResponse{});
+                }
+                dto::SKVRecord k2record(request.collectionName, k2Schema);
+                shd::SKVRecord shdrecord(request.schemaName, shdSchema, std::move(request.key), true);
+                _shdRecToK2(shdrecord, k2record);
+
+                return it->second.handle.read(k2record.getKey(), request.collectionName)
```
jfunston, line 322: Should we use the SKVRecord read instead?

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
iafuture: good point. i'll just pass true

src/k2/httpproxy/HTTPProxy.cpp
iafuture: Yup, makes sense. no idea why i didn't do that in the first place :)

jfunston (approved):

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -395,9 +395,17 @@ HTTPProxy::HTTPProxy() : _client(K23SIClientConfig()) {
...
+    std::vector<seastar::future<>> futs;
+    for (auto& [ts, txn]: _txns) {
+        futs.push_back(txn.handle.kill());
+    }
+    return seastar::when_all_succeed(futs.begin(), futs.end()).discard_result()
+    .then([this] {
+        _txns.clear();
+    })
+    .then([this] {
+        return _client.gracefulStop();
```
ahsank, line 406: What happens if seastar calls `K23SIClient::gracefulStop()` before calling HttpProxy `gracefulStop`? Then _client gracefulstop will be called twice.

ahsank (approved):

src/k2/httpproxy/HTTPProxy.cpp
jfunston: We are not using the K23SIClient as an applet, so gracefulStop is not called for it automatically. Instead the HTTPProxy owns the client directly (it is a member variable) so it is responsible for starting and stopping it.

ahsank: Yes forgot that it's just client api not applet.

PR #313, opened 2022-06-09
Implement http api for end txn

ahsank:


src/skvhttpclient/src/skvhttp/dto/K23SI.h
```diff
@@ -213,13 +213,7 @@ struct TxnEndRequest {
     // the end action (Abort|Commit)
     EndAction action;
 
-    // This flag does not impact correctness, just performance for certain workloads
-    bool syncFinalize=false;
```
jfunston, line 217: We still would like to be able to turn syncFinalize to true. One use case would be in opengauss during the init, we can turn it on to make sure that the cluster is in a steady state before starting a benchmark.

jfunston (changes requested):

src/skvhttpclient/src/skvhttp/dto/K23SI.h
iafuture: agreed. In general, we should have all parameters that the k23si_client library allows

test/integration/skvclient.py
```diff
@@ -97,6 +97,14 @@ def __init__(self, timeout=TimeDelta(seconds=10), priority=TxnPriority.Medium, s
     def serialize(self):
         return [self.timeout.serialize(), self.priority.serialize(), self.syncFinalize]
 
+class EndAction(int, Enum):
```
iafuture, line 100: endAction is already an enum, defined in K23SI.h

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -341,10 +341,38 @@ HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
...
+    return  MakeHTTPResponse<T>(std::move(status), std::move(result));
+}
+
 seastar::future<std::tuple<sh::Status, shd::TxnEndResponse>>
 HTTPProxy::_handleTxnEnd(shd::TxnEndRequest&& request) {
     K2LOG_D(log::httpproxy, "Received txn end request {}", request);
-    return MakeHTTPResponse<shd::TxnEndResponse>(sh::Statuses::S501_Not_Implemented("txn end not implemented"), shd::TxnEndResponse{});
+    auto it = _txns.find(request.timestamp);
+    if (it == _txns.end()) {
+        return makeResponse<shd::TxnEndResponse>(sh::Statuses::S410_Gone("transaction does not exist"));
```
iafuture, line 366: this is a check we do on all requests. Could you please make this status code a constant somewhere to make sure we respond consistently when a txn doesn't exist?

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -341,10 +341,38 @@ HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
...
+    }
+    return it->second.handle.end(request.action == shd::EndAction::Commit)
+        .then([this, &request] (auto&& result) {
+            if (!result.status.is2xxOK()) {
+                return makeResponse<shd::TxnEndResponse>(std::move(result.status));
+            }
+            _txns.erase(request.timestamp);
```
iafuture, line 373: we should make sure we erase the txn from the map on 4xx codes. In other words, weshould only keep it around on a 5xx code (e.g. timeout)

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -341,10 +341,38 @@ HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
     return MakeHTTPResponse<shd::QueryResponse>(sh::Statuses::S501_Not_Implemented("query not implemented"), shd::QueryResponse{});
 }
 
+// Send status with default result
+template <class T>
```
iafuture, line 345: there are already helpers to generate responses, called MakeHTTPResponse. Please use those

iafuture (changes requested):
inline comments added below

src/skvhttpclient/src/skvhttp/dto/K23SI.h
ahsank: `K2TxnHandle::end` function only allows one boolean parameter . So http proxy won't be able to use `syncFinalize` parameter.

```
seastar::future<EndResult> end(bool shouldCommit);
```

https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/module/k23si/client/k23si_client.h#L414

iafuture: You're right! The option we're talking about is passed at begin time (as part of the txn options). This is already handled.
This particular code change removes the flags from the TxnEndRequest. It was left there from the native node API and it doesn't apply to the proxy.

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -341,37 +344,19 @@ HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
...
-
 seastar::future<std::tuple<sh::Status, shd::TxnEndResponse>>
 HTTPProxy::_handleTxnEnd(shd::TxnEndRequest&& request) {
     K2LOG_D(log::httpproxy, "Received txn end request {}", request);
     auto it = _txns.find(request.timestamp);
     if (it == _txns.end()) {
-        return makeResponse<shd::TxnEndResponse>(sh::Statuses::S410_Gone("transaction does not exist"));
+        return MakeHTTPResponse<shd::TxnEndResponse>(Txn_S410_Gone, shd::TxnEndResponse{});
     }
     return it->second.handle.end(request.action == shd::EndAction::Commit)
```
iafuture, line 354: The request should be moved into seastar::do_with() since you're referencing it in the continuation

test/integration/test_http.py
```diff
@@ -88,130 +102,115 @@ def test_basicTxn(self):
...
 
         # Write/Read with bad partition key data type, should fail
-        bad_loc = loc.get_new(partition_key=1)
-        status = txn.write(bad_loc, additional_data)
-        self.assertEqual(status.code, 400)
-        status, _ = txn.read(bad_loc)
-        self.assertEqual(status.code, 400)
+        bad_loc = TestHTTP.schema.make_record(partitionKey=1, rangeKey=b"test1rk", data=b"mydata")
+        status = txn.write(TestHTTP.cname, bad_loc)
+        self.assertEqual(status.code, 503)
```
iafuture, line 159: this is a problem i think - why is the code 503?

test/integration/test_http.py
```diff
@@ -88,130 +102,115 @@ def test_basicTxn(self):
...
-        status = txn.write(loc, additional_data)
-        self.assertEqual(status.code, 201)
-        status, record = txn.read(loc)
-        self.assertEqual(status.code, 200)
-        self.assertEqual(record["data"], "data1")
-        self.assertEqual(record["partitionKey"], "ptest2")
-        self.assertEqual(record["rangeKey"], "rtest2")
+        bad_loc = TestHTTP.schema.make_record(partitionKey=b"test2pk", rangeKey=1, data=1)
+        status = txn.write(TestHTTP.cname, bad_loc)
+        self.assertEqual(status.code, 503)
```
iafuture, line 171: and here

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Good catch! Fixed by copying timestamp.

test/integration/test_http.py
ahsank: Fixed

test/integration/test_http.py
```diff
@@ -88,130 +102,115 @@ def test_basicTxn(self):
         self.assertEqual(resultRec.fields.rangeKey, b"test1rk")
         self.assertEqual(resultRec.fields.data, b"mydata")
 
+        # Commit
+        status = txn.end()
+        self.assertTrue(status.is2xxOK())
+
+        # Commit again, should fail
+        status = txn.end()
+        self.assertFalse(status.is2xxOK())
```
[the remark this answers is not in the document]
ahsank: Fixed

test/integration/test_http.py
```diff
@@ -88,130 +102,115 @@ def test_basicTxn(self):
...
+        status, _ = txn.read(TestHTTP.cname, bad_loc)
+        self.assertEqual(status.code, 503)
 
         # Write/Read with bad range key data type, should fail
-        bad_loc = loc.get_new(range_key=1)
-        status = txn.write(bad_loc, additional_data)
-        self.assertEqual(status.code, 400)
+        bad_loc = TestHTTP.schema.make_record(partitionKey=b"test2pk", rangeKey=1, data=b"mydata")
+        status = txn.write(TestHTTP.cname, bad_loc)
+        self.assertEqual(status.code, 503)
```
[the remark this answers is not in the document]
ahsank: Fixed

iafuture (approved):

jfunston (approved):

PR #315, opened 2022-06-17
Implement Query Api in http proxy

ahsank:


src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -316,7 +234,38 @@ HTTPProxy::_handleRead(shd::ReadRequest&& request) {
...
 HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
     K2LOG_D(log::httpproxy, "Received query request {}", request);
-    return MakeHTTPResponse<shd::QueryResponse>(sh::Statuses::S501_Not_Implemented("query not implemented"), shd::QueryResponse{});
+    auto iter = _txns.find(request.timestamp);
+    if (iter == _txns.end()) {
+        return MakeHTTPResponse<shd::QueryResponse>(Txn_S410_Gone, shd::QueryResponse{});
+    }
+    auto queryIter = iter->second.queries.find(request.queryId);
+    if (queryIter ==iter->second.queries.end()) {
+        return MakeHTTPResponse<shd::QueryResponse>(sh::Statuses::S410_Gone("Query does not exist"), shd::QueryResponse{});
```
iafuture, line 243: We should log a warning here - it signals a bad client state and should not happen since we're the ones keeping track of transactions.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -316,7 +234,38 @@ HTTPProxy::_handleRead(shd::ReadRequest&& request) {
...
+        records.reserve(result.records.size());
+        for (auto& k2record: result.records) {
+            sh::String collectionName(k2record.collectionName);
+            // k2 schema is already populated by query api, get corresponding shd schema from cache
+            auto shSChema = getSchemaFromCache(collectionName, k2record.schema);
+            auto rec = _buildSHDRecord(k2record, collectionName, shSChema);
+            records.push_back(std::move(rec.getStorage()));
+        }
+        auto& txnrec = _txns.at(request.timestamp);
+        auto& query = txnrec.queries.at(request.queryId);
```
iafuture, line 261: these lines can throw an exception. You should probably catch that and convert to some non-retryable error

src/skvhttpclient/src/skvhttp/dto/K23SI.h
```diff
@@ -144,67 +144,25 @@ struct QueryRequest {
...
+    Timestamp timestamp; // identify the issuing transaction
+    uint64_t queryId;
+    K2_SERIALIZABLE_FMT(QueryRequest, timestamp, queryId);
 };
 
-struct CreateQueryResponse {
-    Query query;    // the query if the response was OK
-    K2_SERIALIZABLE_FMT(CreateQueryResponse, query);
+struct QueryResponse {
+    bool done;
```
iafuture, line 163: you should specify initial value of false: {false}

src/skvhttpclient/src/skvhttp/dto/K23SI.h
```diff
@@ -144,67 +144,25 @@ struct QueryRequest {
...
+    K2_SERIALIZABLE_FMT(CreateQueryResponse, queryId);
 };
 
-struct CreateQueryRequest {
-    String collectionName;
-    String schemaName;
-    K2_SERIALIZABLE_FMT(CreateQueryRequest, collectionName, schemaName);
+struct QueryRequest {
+    Timestamp timestamp; // identify the issuing transaction
+    uint64_t queryId;
```
iafuture, line 158: specify initial value of 0: {0}

src/skvhttpclient/src/skvhttp/dto/K23SI.h
```diff
@@ -144,67 +144,25 @@ struct QueryRequest {
...
-
-private:
-    void checkKeysProjected();
-
-    bool done = false;
-    bool inprogress = false;  // Used to prevent user from changing predicates after query has started
-    bool keysProjected = true;
-    QueryRequest request;
+struct CreateQueryResponse {
+    uint64_t queryId;    // the query if the response was OK
```
iafuture, line 152: initial value of 0: {0}

src/skvhttpclient/src/skvhttp/dto/K23SI.cpp
```diff
@@ -22,70 +22,3 @@ Copyright(c) 2020 Futurewei Cloud
 */
 
 #include "K23SI.h"
-namespace skv::http::dto {
```
iafuture, line 25: I guess we can now just delete this file

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -347,10 +296,118 @@ HTTPProxy::_handleGetSchema(shd::GetSchemaRequest&& request) {
     });
 }
 
+void HTTPProxy::setQueryRecord(const sh::String& collectionName, shd::SKVRecord::Storage&& key, dto::SKVRecord& k2record) {
```
iafuture, line 299: This name is confusing to me as I'm not seeing how we're 'setting' a query record anywhere. I'm also not sure what a 'query record' is.
I think this is the equivalent for _shdRecToK2, but with storage. So something like shdStorageToK2Record would be more descriptive.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -347,10 +296,118 @@ HTTPProxy::_handleGetSchema(shd::GetSchemaRequest&& request) {
...
+            if(!result.status.is2xxOK()) {
+                return MakeHTTPResponse<shd::CreateQueryResponse>(sh::Status{.code = result.status.code, .message = result.status.message}, shd::CreateQueryResponse{});
+            }
+            try {
+                setQueryRecord(req.collectionName, std::move(req.key), result.query.startScanRecord);
+                setQueryRecord(req.collectionName, std::move(req.endKey), result.query.endScanRecord);
+
+                if (req.recordLimit >= 0) result.query.setLimit(req.recordLimit);
+                if (req.includeVersionMismatch) result.query.setIncludeVersionMismatch(req.includeVersionMismatch);
+                if (req.reverseDirection) result.query.setReverseDirection(req.reverseDirection);
```
iafuture, line 363: i don't think we need to check these values - just set them on the result.query

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -347,10 +296,118 @@ HTTPProxy::_handleGetSchema(shd::GetSchemaRequest&& request) {
...
 }
 
+void HTTPProxy::setQueryRecord(const sh::String& collectionName, shd::SKVRecord::Storage&& key, dto::SKVRecord& k2record) {
+    if (key.fieldData.size() == 0) return;
+    auto shdSchema = getSchemaFromCache(collectionName, k2record.schema);
+    shd::SKVRecord shdrecord(collectionName, shdSchema, std::move(key), true);
+    _shdRecToK2(shdrecord, k2record);
+}
+
+namespace dtoexp = dto::expression;
```
iafuture, line 306: a 'k2exp' would be better here since we want to distinguish k2 vs sh expressions

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -347,10 +296,118 @@ HTTPProxy::_handleGetSchema(shd::GetSchemaRequest&& request) {
...
+                        projection.push_back(String(std::move(p)));
+                    }
+                    result.query.addProjection(projection);
+                }
+            } catch(shd::DeserializationError& err) {
+                return MakeHTTPResponse<shd::CreateQueryResponse>(sh::Statuses::S400_Bad_Request(err.what()), shd::CreateQueryResponse{});
+            }
+
+            auto queryId = _queryID++;
+            _txns.at(req.timestamp).queries[queryId] = std::move(result.query);
```
iafuture, line 381: this can throw.

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: It cannot throw as previously in line 349 it checks if timestamp txn exists.

src/skvhttpclient/src/skvhttp/dto/Expression.h
```diff
@@ -78,6 +78,23 @@ struct Value {
 
     K2_SERIALIZABLE(Value, fieldName, type, literal);
 
+   // Extracts a given type for this Value.
```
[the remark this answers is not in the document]
ahsank: Done

test/integration/skvclient.py
```diff
@@ -135,20 +192,43 @@ def read(self, collection, keyrec) :
...
+    def query(self, query):
+        status, result = self.client.make_call("/api/Query", [self.timestamp, query.query_id])
+        if not status.is2xxOK():
+            return status, None, None
+        records = []
+        for storage in result[1]:
+            status, schema = self.client.get_schema(query.cname, query.sname, storage[2])
+            if not status.is2xxOK():
+                return status, None, None
+            record = schema.parse_read(storage, None)
```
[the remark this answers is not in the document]
ahsank: Done

src/k2/httpproxy/HTTPProxy.cpp

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
iafuture: yes it can. Same reason as other instances

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -379,7 +388,11 @@ HTTPProxy::_handleCreateQuery(shd::CreateQueryRequest&& request) {
             }
 
             auto queryId = _queryID++;
-            _txns.at(req.timestamp).queries[queryId] = std::move(result.query);
+            try {
```
iafuture, line 391: This works, but the rest of our codebase uses the iterator pattern instead of exception-based checks for existence. Please use iterator check instead:
`
if (auto it = _txns.find(req.timestamp); it == _txns.end()) {
   // send error
}
else {
   // use *it
}
`

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -258,14 +259,22 @@ HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
             auto rec = _buildSHDRecord(k2record, collectionName, shSChema);
             records.push_back(std::move(rec.getStorage()));
         }
-        auto& txnrec = _txns.at(request.timestamp);
-        auto& query = txnrec.queries.at(request.queryId);
-        bool isDone = query.isDone();
-        if (isDone) {
-            txnrec.queries.erase(request.queryId);
+        try {
```
iafuture, line 262: please use iterator pattern as described below

iafuture (changes requested):

iafuture (approved):

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -316,7 +235,45 @@ HTTPProxy::_handleRead(shd::ReadRequest&& request) {
...
+        } else {
+            if (auto queryIter = iter->second.queries.find(request.queryId); queryIter ==iter->second.queries.end()) {
+                K2LOG_W(log::httpproxy, "Query not found, txn: {} query: {}", request.timestamp, request.queryId);
+                return MakeHTTPResponse<shd::QueryResponse>(Query_S410_Gone, shd::QueryResponse{});
+            } else {
+                if (queryIter->second.isDone()) {
+                    iter->second.queries.erase(request.queryId);
+                }
+                return MakeHTTPResponse<shd::QueryResponse>(sh::Status{.code=result.status.code, .message=result.status.message},
+                    shd::QueryResponse{.done = queryIter->second.isDone(), .records = std::move(records)});
```
jfunston, line 273: Is it safe to access queryIter here if the erase happens above?

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Good catch! I mistakenly refactored this way in previous update to reduce a line of code.

jfunston (approved):

PR #316, opened 2022-06-20
Implement basic operations in skvhttp client. 

jfunston:
Query and partial update are still unimplemented

### DIFF
--- a/src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
+++ b/src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
@@ -1,0 +1,157 @@
+/*
+MIT License
+
+Copyright(c) 2022 Futurewei Cloud
+
+    Permission is hereby granted,
+    free of charge, to any person obtaining a copy of this software and associated documentation files(the "Software"), to deal in the Software without restriction, including without limitation the rights to use, copy, modify, merge, publish, distribute, sublicense, and / or sell copies of the Software, and to permit persons to whom the Software is furnished to do so, subject to the following conditions :
+
+    The above copyright notice and this permission notice shall be included in all copies
+    or
+    substantial portions of the Software.
+
+    THE SOFTWARE IS PROVIDED "AS IS",
+    WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+    FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT.IN NO EVENT SHALL THE
+    AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM,
+    DAMAGES OR OTHER
+    LIABILITY,
+    WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+*/
+
+#include "SKVClient.h"
+#include "skvhttp/common/Common.h"
+#include "skvhttp/common/Status.h"
+#include "skvhttp/dto/ControlPlaneOracle.h"
+#include "skvhttp/dto/K23SI.h"
+#include "skvhttp/dto/SKVRecord.h"
+#include <cstdint>
+#include <memory>
+#include <netinet/ip.h>
+#include <unordered_map>
+#include <vector>
+
+using namespace skv::http;
+
+boost::future<Response<>> Client::createCollection(dto::CollectionMetadata metadata, std::vector<String> rangeEnds) {
+  dto::CollectionCreateRequest request {
+    .metadata = std::move(metadata),
+    .rangeEnds = std::move(rangeEnds)
+  };
+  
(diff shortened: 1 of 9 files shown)


src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -0,0 +1,157 @@
...
+    OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+    SOFTWARE.
+*/
+
+#include "SKVClient.h"
+#include "skvhttp/common/Common.h"
+#include "skvhttp/common/Status.h"
+#include "skvhttp/dto/ControlPlaneOracle.h"
+#include "skvhttp/dto/K23SI.h"
+#include "skvhttp/dto/SKVRecord.h"
```
iafuture, line 29: include via <>?

src/skvhttpclient/src/skvhttp/dto/SKVRecord.h
```diff
@@ -229,7 +229,8 @@ class SKVRecordBuilder {
             }
         }
 
-        _writer.write(field);
```
iafuture, line 232: I don't think this is right - the value should be written as-is. We read it as T on the reverse path, not optional<T>

src/skvhttpclient/src/skvhttp/client/SKVClient.h
```diff
@@ -22,13 +22,17 @@ Copyright(c) 2022 Futurewei Cloud
 */
 
 #pragma once
+#include "skvhttp/common/Serialization.h"
```
iafuture, line 25: <> include?

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -0,0 +1,157 @@
...
+  dto::GetSchemaRequest request {
+    .collectionName = collectionName,
+    .schemaName = schemaName,
+    .schemaVersion = schemaVersion
+  };
+
+  return _HTTPClient.POST<dto::GetSchemaRequest, dto::GetSchemaResponse>("/api/GetSchema", std::move(request)).
+    then([this, collectionName, schemaName, schemaVersion] (auto&& futResp) {
+      auto&& [status, resp] = futResp.get();
+      auto schema = std::make_shared<dto::Schema>(resp.schema);
```
iafuture, line 83: you should probably only do this on 2xxOK(). And return nullptr otherwise

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+   return _HTTPClient.POST<dto::TxnBeginRequest, dto::TxnBeginResponse>("/api/TxnBegin", std::move(request)).
+    then([this] (auto&& futResp) {
+      auto&& [status, resp] = futResp.get();
+      TxnHandle txn(this, resp.timestamp);
+      return Response<TxnHandle>(std::move(status), std::move(txn));
+    });
+}
+
+boost::future<Response<>> TxnHandle::write(dto::SKVRecord& record, bool erase, dto::ExistencePrecondition precondition) {
```
iafuture, line 104: will a partial update be a separate API? if not, maybe just take the vector of fields as a param here

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -0,0 +1,157 @@
...
+    .value = record.storage.share(),
+    .fieldsForPartialUpdate = std::vector<uint32_t>{}
+  };
+
+  return _client->_HTTPClient.POST<dto::WriteRequest>("/api/Write", std::move(request));
+}
+
+boost::future<Response<dto::SKVRecord>> TxnHandle::read(dto::SKVRecord record) {
+  if (record.fieldCursor != record.schema->fields.size()) {
+    return MakeResponse(Statuses::S400_Bad_Request("All fields of record must be serialized for read request"), dto::SKVRecord{});
```
iafuture, line 120: really? I thought only keys are needed

iafuture (changes requested):

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
jfunston: The problem is the proxy tries to deserialize all the fields to make the storage object on its side, and we can't fix it here for the user since we need the builder for that. We could take a builder as the parameter here but then the user can't reuse records or use reconstructed records like for secondary index usage.

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
jfunston: yeah, the plan is to have partial update separate

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
jfunston: good catch, emacs is being very aggressive about adding its own includes

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
iafuture: i see. i think it's fine to leave as-is

iafuture (approved):

PR #318, opened 2022-06-22
Expire idle txns in http proxy

ahsank:


src/k2/common/ExpiryList.h
```diff
@@ -0,0 +1,85 @@
...
+
+    void add(ElemT& elem) {
+        _list.push_back(elem);
+    }
+
+
+private:
+    typedef nsbi::list<ElemT, nsbi::member_hook<ElemT, nsbi::list_member_hook<>, Field>> IntrusiveList;
+    IntrusiveList _list;
+    // heartbeats checks are driven off single timer.
```
iafuture, line 82: what are "heartbeats" ?

test/integration/test_http.py
```diff
@@ -465,6 +465,31 @@ def test_query(self):
         status = txn.end()
         self.assertTrue(status.is2xxOK())
 
+    def test_txn_timeout(self):
+        # Begin Txn with timeout 1s
```
iafuture, line 469: could we make these timeouts tighter so we don't have to wait so much for the entire test to complete?

src/k2/httpproxy/HTTPProxy.h
```diff
@@ -109,6 +119,9 @@ class HTTPProxy {
         std::unordered_map<String,
             std::unordered_map<uint32_t, std::shared_ptr<shd::Schema>>
     >> _shdSchemas;
+    ExpiryList<ManagedTxn, &ManagedTxn::tsLink> _expiryList;
+    // Txn idle timeout
+    ConfigDuration httpproxy_txn_timeout{"httpproxy_txn_timeout", 60s};
```
iafuture, line 124: please use camelCase for variables, not underscores. Also, this is a private variable so it should start with "_"

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -441,6 +450,7 @@ HTTPProxy::HTTPProxy() : _client(K23SIClientConfig()) {
 
 seastar::future<> HTTPProxy::gracefulStop() {
     std::vector<seastar::future<>> futs;
+    _expiryList.stop().wait();
```
iafuture, line 453: calling wait or get should not be common practice as these block.just push it to the futures: futs.push_back(_expiryList.stop())

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -456,6 +466,15 @@ seastar::future<> HTTPProxy::gracefulStop() {
 seastar::future<> HTTPProxy::start() {
     _registerMetrics();
     _registerAPI();
+    _expiryList.start(httpproxy_txn_timeout(), [this](ManagedTxn& txn) {
+        auto ts = txn.timestamp;
+        K2LOG_I(log::httpproxy, "Removing txn {} because of timeout", ts);
+        return txn.handle.kill()
+            .then([this, ts]{
```
iafuture, line 473: I think there maybe a range of race conditions with the list and _txns here. I think we'd have to move the handle out of _txns (i.e. extract() it from the map) and then call kill on it

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -262,6 +266,7 @@ HTTPProxy::_handleQuery(shd::QueryRequest&& request) {
         if (auto iter = _txns.find(request.timestamp); iter == _txns.end()) {
             return MakeHTTPResponse<shd::QueryResponse>(Txn_S410_Gone, shd::QueryResponse{});
         } else {
+            updateExpiry(iter->second);
```
iafuture, line 269: why update expiry here? We already updated when we received query

iafuture (changes requested):

src/k2/common/ExpiryList.h
```diff
@@ -0,0 +1,85 @@
...
+        return _expiryTimer.stop()
+            .then([this] {
+                _list.clear();
+             });
+    }
+
+    // Move elment to the end of the TS ordered list
+    void moveLast(ElemT& elem) {
+        // new TS needs to be >= current maximum TS
+        if (elem.expiry() < _list.back().expiry()) {
```
iafuture, line 62: why this requirement at all? I think it's fine to expire later than all other entries

src/k2/common/ExpiryList.h
```diff
@@ -0,0 +1,85 @@
...
+            assert(false);
```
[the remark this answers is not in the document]
ahsank: Done

src/k2/common/ExpiryList.h
```diff
@@ -0,0 +1,85 @@
...
+
+namespace k2 {
+namespace nsbi = boost::intrusive;
+
+template <class ElemT, nsbi::list_member_hook<> ElemT::* Field, class ClockT=Clock>
+class ExpiryList {
+public:
+    typedef std::function<seastar::future<>(ElemT&)> Func;
+
+    void start(Duration timeout, Func&& cleanupFn) {
```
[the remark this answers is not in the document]
ahsank: Added comment.

src/k2/common/ExpiryList.h
```diff
@@ -0,0 +1,85 @@
...
+
+#pragma once
+#include <boost/intrusive/list.hpp>
+#include "Timer.h"
+
+namespace k2 {
+namespace nsbi = boost::intrusive;
+
+template <class ElemT, nsbi::list_member_hook<> ElemT::* Field, class ClockT=Clock>
+class ExpiryList {
```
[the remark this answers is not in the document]
ahsank: Added comment.

src/k2/common/ExpiryList.h
ahsank: If the list is not ordered by expiry time and front of the list has large expiry time in future, late elements may not get expired.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: If there are large number of pages in query, the client may take long time to retrieve data. During time the txn should be kept alive.

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Fixed.

src/k2/common/ExpiryList.h
```diff
@@ -0,0 +1,85 @@
...
+
+    seastar::future<> stop() {
+        return _expiryTimer.stop()
+            .then([this] {
+                _list.clear();
+             });
+    }
+
+    // Move elment to the end of the TS ordered list
+    void moveLast(ElemT& elem) {
```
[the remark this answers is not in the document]
ahsank: Fixed.

src/k2/common/Log.h
```diff
@@ -0,0 +1,29 @@
+/*
+MIT License
+
+Copyright(c) 2021 Futurewei Cloud
```
iafuture, line 4: copyright 22

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -456,6 +466,16 @@ seastar::future<> HTTPProxy::gracefulStop() {
 seastar::future<> HTTPProxy::start() {
     _registerMetrics();
     _registerAPI();
+    // Run timer at interval minimum of 1s and txn_timeout/2.
+    // After expiry it may take additional min(1s,timeout/2) for cleanup.
+    auto interval = _txnTimeout()/2;
+    if (interval > 1s) interval = 1s;
```
iafuture, line 472: comment says min 1s, but here you're capping it at 1s (i.e. the max). Did you mean to compare < ? Also, this check is probably not needed - we're configuring this on the cmdline anyway. You can log an Info message to say what interval is used

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -456,6 +466,16 @@ seastar::future<> HTTPProxy::gracefulStop() {
...
+    _expiryList.start(interval, [this](ManagedTxn& txn) {
+        auto ts = txn.timestamp;
+        K2LOG_I(log::httpproxy, "Removing txn {} because of timeout", ts);
```
iafuture, line 475: Probably make this a warning  i think.

iafuture (changes requested):

src/k2/common/Log.h
ahsank: Fixed

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Fixed

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Changed to use config value instead of trying to calculate interval based on txn timeout.

iafuture (dismissed):

jfunston (dismissed):

ahsank:
Did git rebase

PR #320, opened 2022-06-27
Remove test sleep

ccjeff:
There are a couple of changes made along with removing test sleeps:
1. adding retry operations for the TSOClient so that it keeps sending get tso requests to the CPO; enables async startup of TSOs and TSOClients.
2. Exponential Backoff strategy now sleeps before execute the next retry
3. Adding retry partition assignment in the CPO (to allow async startup of nodepools).
4. Using successful get timestamp using the TSO client as the signal for starting the tests. (remove sleep time)
5. Changed the inconsistent configuration variable naming (cpo_endpoint -> cpo)
6. Updates the 3sitxn test to use batch generated timestamps instead of manually create the timestamps. (The interval might be too large, depending on the execution speed the previous test's timestamp can cause problem for the next test)

### DIFF
--- a/src/k2/assignmentManager/AssignmentManager.cpp
+++ b/src/k2/assignmentManager/AssignmentManager.cpp
@@ -70,6 +70,13 @@ AssignmentManager::handleAssign(dto::AssignmentCreateRequest&& request) {
     dto::Partition& partition = request.partition;
 
     if (_pmodule) {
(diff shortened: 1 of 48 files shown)


ccjeff:
Just rebased from master and tested. Everything still works :)

ccjeff:
Changes now in new branch

PR #323, opened 2022-06-28
Add query implementation for the skvhttpclient

jfunston:


test/integration/skvclient.py
```diff
@@ -300,6 +300,18 @@ def make_record(self, **dataFields):
...
+    # For use in creating key records for a prefix query. The difference is fields that are not set
+    # are not set in excludedFields and are not set to None
+    def make_prefix_record(self, **dataFields):
+        rec = Record(self.name, self.version)
+        for i,field in enumerate(self.fields):
+            dname = field.name.decode('ascii')
+            if dname in dataFields:
+                fieldValue = dataFields.get(dname)
+                rec._posFields.append(fieldValue)
+                rec.fields.__dict__[dname] = fieldValue
```
iafuture, line 312: i think in the else part we should break here to make sure we don't set any other fields. Maybe throw after if there are some dataFields left that we didn't use (just check if len(dataFields) == len(rec._posFields))

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -38,22 +38,26 @@ static const inline auto Txn_S410_Gone = sh::Statuses::S410_Gone("transaction do
...
-    shdrec.visitRemainingFields([&k2rec](const auto&, auto&& value) mutable {
-        if (value) {
-            using T = typename std::remove_reference_t<decltype(value)>::value_type;
-            if constexpr (std::is_same_v<T, shd::FieldType>) {
-                auto kval = static_cast<dto::FieldType>(to_integral(*value));
-                k2rec.serializeNext<dto::FieldType>(kval);
-            } else if constexpr (std::is_same_v<T, sh::String>) {
-                String k2str(value->data(), value->size());
-                k2rec.serializeNext<String>(std::move(k2str));
+    uint32_t serializedCursor = shdrec.storage.serializedCursor;
```
iafuture, line 41: Just to be safe, i think we should seek(0) on shdrec

iafuture (changes requested):

src/skvhttpclient/src/skvhttp/dto/SKVRecord.h
```diff
@@ -130,12 +130,15 @@ class SKVRecord {
     struct Storage {
         // Bitmap of fields that are excluded because they are optional or this is for a partial update
         std::vector<bool> excludedFields;
+        // This is different from excludedFields above and needed because some use cases of the SKVRecord
+        // (e.g. specifying a prefix scan for queries) will not completely serialize the record
+        uint32_t serializedCursor = 0;
```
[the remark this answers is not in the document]
jfunston: It is tricky because with the query we allow prefix scans of the key fields, which is expressed by the user as a partially created record, which is different from serializing null. Before the proxy expected a fully created record when it converted between the skvclient record and a k2 record. With this change the proxy knows how many fields were actually serialized by the user and only copies those fields to the k2 record.

iafuture (approved):

PR #326, opened 2022-06-29
Support partial update in http proxy and python client

ahsank:


ahsank:
Added test to update range key using partial update. having unexpected behavior, the data has new range key but indexed by old range key.

```python
#  Update range key using partial update, old range key is test1rk
record = TestHTTP.schema.make_record(rangeKey=b"test1rk_update")
record.fieldsForPartialUpdate = [b"rangeKey"]
key = TestHTTP.schema.make_record(partitionKey=b"test1pk", rangeKey=b"test1rk")
status = txn.write(TestHTTP.cname, record, key=key)
self.assertTrue(status.is2xxOK())

# read using old range key, should fail but succeeds and data seems to be
# indexed by old range key but, it's value is new range key
status, resultRec = txn.read(TestHTTP.cname, key)
self.assertTrue(status.is2xxOK())
self.assertEqual(resultRec.fields.partitionKey, b"test1pk")
self.assertEqual(resultRec.fields.rangeKey, b"test1rk_update")
```

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -167,6 +167,34 @@ HTTPProxy::_handleTxnBegin(shd::TxnBeginRequest&& request){
...
+        _shdRecToK2(shdKeyRecord, k2KeyRecord);
+    }  catch(shd::DeserializationError& err) {
+        return MakeHTTPResponse<shd::WriteResponse>(sh::Statuses::S400_Bad_Request(err.what()), shd::WriteResponse{});
+    }
+    std::vector<String> fieldsForPartialUpdate;
+    fieldsForPartialUpdate.reserve(request.fieldsForPartialUpdate.size());
+    for (sh::String& field: request.fieldsForPartialUpdate) {
+        fieldsForPartialUpdate.push_back(String(std::move(field)));
+    }
+    return txn.partialUpdate(k2record, fieldsForPartialUpdate, k2KeyRecord.getKey())
```
jfunston, line 192: We are trying to avoid using the interface with the key specified directly, so you should only pass a record and the fieldsforupdate. The record's key fields will be used to specify which key to update. This also avoids the problem you mentioned where the key fields can be changed.

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -101,7 +101,8 @@ boost::future<Response<>> TxnHandle::write(dto::SKVRecord& record, bool erase, d
         .isDelete = erase,
         .precondition = precondition,
         .value = record.storage.share(),
-        .fieldsForPartialUpdate = std::vector<uint32_t>{}
+        .fieldsForPartialUpdate = std::vector<String>{},
+        .key = dto::SKVRecord::Storage{}
```
jfunston, line 105: You can remove the key from here so that we don't give the user an option to specify a separate key and value record as mentioned above.

test/integration/test_http.py
```diff
@@ -129,6 +129,54 @@ def test_basicTxn(self):
...
+
+        # Write partial upate
+        record = TestHTTP.schema.make_record(rangeKey=b"test1rk_update")
+        record.fieldsForPartialUpdate = [b"rangeKey"]
+        key = TestHTTP.schema.make_record(partitionKey=b"test1pk", rangeKey=b"test1rk")
+        status = txn.write(TestHTTP.cname, record, key=key)
+        self.assertTrue(status.is2xxOK())
+
+        # read using old range key, should fail but succeeds as data seems to be
+        # indexed by old range key but, it's value is new range key
```
jfunston, line 168: We can disallow this for the proxy as mentioned above but this is a good find since the k23si client should check for this. I will add an issue for it.

jfunston (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
ahsank: Changed key field and using key from record.

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
ahsank: Done.

test/integration/test_http.py
ahsank: Done.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -167,6 +186,23 @@ HTTPProxy::_handleTxnBegin(shd::TxnBeginRequest&& request){
...
+    return txn.write(k2record, request.isDelete, static_cast<dto::ExistencePrecondition>(request.precondition))
+        .then([](WriteResult&& result) {
+            return MakeHTTPResponse<shd::WriteResponse>(sh::Status{.code = result.status.code, .message = result.status.message}, shd::WriteResponse{});
+        });
+}
+
+seastar::future<std::tuple<sh::Status, shd::WriteResponse>>
+HTTPProxy::_handlePartialUpdate(K2TxnHandle& txn, shd::WriteRequest&& request, dto::SKVRecord&& k2record) {
+    std::vector<String> fieldsForPartialUpdate = shVectorToK2<String>(request.fieldsForPartialUpdate);
+    return txn.partialUpdate(k2record, fieldsForPartialUpdate, k2record.getKey())
```
jfunston, line 200: You can remove the third parameter here just to make it a bit clearer, partialUpdate has a default defined for it.

jfunston (changes requested):

jfunston (dismissed):

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -79,6 +79,25 @@ void _buildSHDRecordHelperVisitor(std::optional<T> value, String&, shd::SKVRecor
     }
 }
 
+// Convert skv element vector to k2 element vector by convert function
+template <class K2Type, class SHType, typename Fn>
+auto shVectorToK2(std::vector<SHType>&  shElems, Fn convertFn) {
```
iafuture, line 84: These two helper functions end up consuming the elements in the vector. You should instead require the caller to call you with rvalue ref (i.e. shVectorToK2(std::move(myvec))

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -167,6 +186,23 @@ HTTPProxy::_handleTxnBegin(shd::TxnBeginRequest&& request){
...
+    return txn.write(k2record, request.isDelete, static_cast<dto::ExistencePrecondition>(request.precondition))
+        .then([](WriteResult&& result) {
+            return MakeHTTPResponse<shd::WriteResponse>(sh::Status{.code = result.status.code, .message = result.status.message}, shd::WriteResponse{});
+        });
+}
+
+seastar::future<std::tuple<sh::Status, shd::WriteResponse>>
+HTTPProxy::_handlePartialUpdate(K2TxnHandle& txn, shd::WriteRequest&& request, dto::SKVRecord&& k2record) {
+    std::vector<String> fieldsForPartialUpdate = shVectorToK2<String>(request.fieldsForPartialUpdate);
+    return txn.partialUpdate(k2record, fieldsForPartialUpdate)
```
iafuture, line 200: you should move the fieldsForPartialUpdate here

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -101,7 +101,7 @@ boost::future<Response<>> TxnHandle::write(dto::SKVRecord& record, bool erase, d
         .isDelete = erase,
         .precondition = precondition,
         .value = record.storage.share(),
-        .fieldsForPartialUpdate = std::vector<uint32_t>{}
+        .fieldsForPartialUpdate = std::vector<String>{},
```
iafuture, line 104: What is the reason  you want to use strings instead of indices? The k23si txn would translate these strings to indices anyway so might as well just pass the indices I think.

iafuture (changes requested):

src/k2/httpproxy/HTTPProxy.cpp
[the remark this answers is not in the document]
ahsank: Fixed.

src/k2/httpproxy/HTTPProxy.cpp
```diff
@@ -79,6 +79,25 @@ void _buildSHDRecordHelperVisitor(std::optional<T> value, String&, shd::SKVRecor
...
+    k2Elems.reserve(shElems.size());
+    for (SHType& elem: shElems) {
+        k2Elems.push_back(convertFn(std::move(elem)));
+    }
+    return k2Elems;
+}
+
+// Convert skv elment vector to k2 element vector using default conversion
+template <class K2Type, class SHType>
+auto shVectorToK2(std::vector<SHType>&  shElems) {
```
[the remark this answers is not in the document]
ahsank: Fixed.

src/k2/httpproxy/HTTPProxy.cpp

src/k2/httpproxy/HTTPProxy.cpp

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
ahsank: I missed that there is a client Api that takes field indices. Switched to using indices.

PR #335, opened 2022-07-20
Refactored cpoclient to check for the assignment status for all partitions

ccjeff:
Refactored the cpoclient;
Adding checks to allow assignment status check for all partitions (instead of the first partition in the previous version);
Removed partition assignment sleeps from tests;
Changed `CachedClock::Now` to be true by default


src/k2/cpo/client/Client.h
```diff
@@ -69,14 +69,27 @@ class CPOClient {
...
+                        return seastar::make_ready_future<Status>(status);
+                    });
                 });
             }
 
             return seastar::make_ready_future<Status>(std::move(status));
         });
     }
 
+
```
iafuture, line 86: no need for this public method. i think we can just make _getPartitionGetterForCollection public and rename it

test/cpo/CMakeLists.txt
```diff
@@ -1,6 +1,8 @@
 file(GLOB HEADERS "*.h")
 file(GLOB SOURCES "*.cpp")
 
-add_executable (cpo_test ${HEADERS} ${SOURCES})
+add_executable (cpo_test ${HEADERS} CPOTest.cpp Main.cpp)
+add_executable(cpo_client_test CPOClientTest.cpp)
 
 target_link_libraries (cpo_test PRIVATE appbase Seastar::seastar dto tso_client)
+target_link_libraries (cpo_client_test PRIVATE appbase Seastar::seastar dto tso_client cpo_client transport common)
```
iafuture, line 8: missing new line at end of file. You should auto-enable this in your editor settings

test/k23si/K23SITest.cpp
```diff
@@ -587,6 +583,7 @@ seastar::future<> runScenario05() {
 int main(int argc, char** argv) {
     k2::App app("K23SITest");
     app.addOptions()("cpo", bpo::value<k2::String>(), "The endpoint of the CPO");
+    app.addOptions()("create_ddl", bpo::value<k2::ParseableDuration>(), "The duration to wait after called create partition");
```
iafuture, line 586: I'd describe this as "the duration to wait for collection to be created". Also an option that sounds like a command is very confusing for what it is expecting(duration). Something like: "cpo_timeout" or "cpo_create_collection_timeout" would be better

test/cpo/CPOClientTest.cpp
```diff
@@ -0,0 +1,147 @@
...
+        seastar::future<> runTest1() {
+            K2LOG_I(k2::log::cpo_client_test, ">>> Test 1: Getting a non-existent collection");
+            return _cpoClient.getPartitionGetterWithRetry(Deadline<>(100ms),String(collname))
+            .then([this] (auto&& response) {
+                auto&[status, resp] = response;
+                K2EXPECT(k2::log::cpo_client_test, status.is4xxNonRetryable(), true);
+            });
+        }
+        seastar::future<> runTest2() {
+            K2LOG_I(k2::log::cpo_client_test, ">>> Test 2: Creating a collection");
```
iafuture, line 65: this test is expecting to fail, but it doesn't say why. Could you mention somewhere why this should fail?

src/k2/cpo/client/Client.h
```diff
@@ -236,6 +249,40 @@ class CPOClient {
         });
     }
 
+    // Check if all partitions are assigned within the deadline. Returns a future of assignment status.
+    template<typename ClockT=Clock>
+    seastar::future<Status> _checkAllParititonsAssigned(Deadline<ClockT> deadline, const String& name, uint8_t retries=5) {
+        if (!retries) {
+            return seastar::make_ready_future<Status>(Statuses::S408_Request_Timeout("check partition assigned retries exceeded"));
+        }
+        return _getAssignedPartitionWithRetry(deadline, name, dto::Key{})
```
iafuture, line 258: this is a problem i think for the way the helper _getAssignedPartitionWithRetry works. It uses a cached response if partition 0 has been assigned

iafuture (changes requested):

test/cpo/CMakeLists.txt
ccjeff: This is weird. I've had it turned on in my editor and have checked manually that there is an ending new line in my local file.

src/k2/cpo/client/Client.h
```diff
@@ -261,30 +314,102 @@ class CPOClient {
...
+                return seastar::make_ready_future<Status>(std::move(status));
+            }
+        });
+    }
+
+    template<typename ClockT=Clock>
+    seastar::future<Status> _refreshCollectionWithRetry(Deadline<ClockT> deadline, const String& name, bool reverse=false, bool excludedKey=false) {
+        // Always make the remote calls and queue waiters to force a refresh of the collection, retry until deadline runs out
+        // Register the ongoing request
+        K2ASSERT(log::cpoclient, _requestWaiters[name].empty(), "_requestWaiters[name] is not empty");
```
iafuture, line 401: log the name which caused the problem.

src/k2/cpo/client/Client.h
```diff
@@ -202,54 +223,86 @@ class CPOClient {
...
+                    return seastar::make_ready_future<Status>(Statuses::S200_OK("Collection created"));
+                }
+            } 
+            return seastar::make_ready_future<Status>(std::move(status));
+        });
+    }
+
+    // Check if the collection is in cache or the partition is not assigned
+    template<typename ClockT=Clock>
+    seastar::future<Status> _partitionGetterHelper(Deadline<ClockT> deadline, const String& cname, const dto::Key& key=dto::Key{},
```
iafuture, line 262: I think we should rewrite this a bit to make it cleaner, e.g.
```c++
        // If collection is not in cache or partition is not assigned, get collection first
        if (auto it = collections.find(cname); it != collections.end()) {
            K2LOG_D(log::cpoclient, "Collection {} found in cache", cname);

            // we have it but it is not in a good state - get from CPO
            dto::Partition* partition = it->second->getPartitionForKey(key, reverse, exclusiveKey).partition;
            if (!partition || partition->astate != dto::AssignmentState::Assigned) {
                K2LOG_D(log::cpoclient, "Collection {} found but is in bad state {}", cname, partition);
                return _getAssignedPartitionWithRetry(deadline, cname, key, reverse, exclusiveKey);
            }
        }
        else {
            K2LOG_D(log::cpoclient, "Collection {} not found in cache", cname);
            // we don't have the collection - get it from the CPO
            return _getAssignedPartitionWithRetry(deadline, cname, key, reverse, exclusiveKey);
        }

        return seastar::make_ready_future<Status>(Statuses::S200_OK("default cached response"));
```

iafuture (changes requested):

ccjeff:
Closing this due to merging issue.

PR #338, opened 2022-08-05
Updated cpoclient to check for the assignment status for all partitions

ccjeff:
See #335 
Refactored the cpoclient;
Adding checks to allow assignment status check for all partitions (instead of the first partition in the previous version);
Removed partition assignment sleeps from tests;
Changed CachedClock::Now to be true by default

### DIFF
--- a/src/k2/cpo/client/Client.h
+++ b/src/k2/cpo/client/Client.h
@@ -101,7 +101,7 @@ public:
     getPartitionGetterWithRetry(Deadline<ClockT> deadline, const String& cname, const dto::Key& key=dto::Key{},
                                     bool reverse=false, bool exclusiveKey=false) {
         return _partitionGetterHelper(deadline, cname, key, reverse, exclusiveKey)
-        .then([this, &cname] (auto&& status) {
+        .then([this, cname] (auto&& status) {
             auto it = collections.find(cname);
             if (it == collections.end()) {
                 // Failed to get collection, returning status from GetAssignedPartitionWithRetry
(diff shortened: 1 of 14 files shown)


src/k2/transport/RetryStrategy.h
```diff
@@ -97,13 +97,11 @@ class ExponentialBackoffStrategy {
...
                         (*resultPtr.get()) = std::move(fut);
-                        K2LOG_D(log::txretry, "round {} ended with success={}", _try, _success);
-                        if (!_success) {
-                            _try++;
-                            _currentTimeout = Duration((uint64_t)(nsec(_currentTimeout).count() * _rate));
-                            if (!deadline.isOver()) {
-                                return seastar::sleep(deadline.getRemaining());
-                            }
+                        K2LOG_D(log::tx, "round ended with success={}", _success);
+                        if (!_success && !deadline.isOver()) {
```
jfunston, line 101: Looks like this is reverting Ivan's previous fix. Is that intentional or left over from the merge problems?

test/k23si/K23SITest.cpp
```diff
@@ -90,15 +90,11 @@ class K23SITest {
                     },
                     .rangeEnds{}
                 };
-                return RPC().callRPC<dto::CollectionCreateRequest, dto::CollectionCreateResponse>
-                        (dto::Verbs::CPO_COLLECTION_CREATE, request, *_cpoEndpoint, 1s);
+                return _cpo_client.createAndWaitForCollection(Deadline<>(_createWaitTime()), std::move(request.metadata), std::move(request.rangeEnds));
             })
-            .then([](auto&& response) {
+            .then([](Status&& status) {
```
jfunston, line 95: Seems like this .then block is a no-op now and the K2EXPECT call can be moved into the next one.

jfunston (changes requested):

src/k2/transport/RetryStrategy.h
ccjeff: It looks like a merging problem. I'll fix manually.

iafuture (changes requested):

src/k2/cpo/client/Client.h
```diff
@@ -303,8 +418,8 @@ class CPOClient {
                 return seastar::make_ready_future<Status>(std::move(status));
             }
 
-            if (status.is2xxOK() && retry && !retries) {
-                K2LOG_D(log::cpoclient, "not all partitions have been assigned in cpo yet");
+            if (status.is2xxOK() && retry) {
```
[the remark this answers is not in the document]
ccjeff: hmm I don't think I get why we need to check for remaining deadline here. I think retry checks for both failed collection get and unassigned or missing partitions.

src/k2/cpo/client/Client.h
```diff
@@ -202,54 +220,76 @@ class CPOClient {
...
+
+    // Check if all partitions are assigned within the deadline. Returns a future of assignment status.
+    template<typename ClockT=Clock>
+    seastar::future<Status> _checkAllParititonsAssigned(Deadline<ClockT> deadline, const String& name) {
+        return _getAssignedPartitionWithRetry(deadline, name, dto::Key{})
+        .then([this, name, deadline](Status&& status){
+            K2LOG_D(log::cpoclient, "Checking if all partitions are assigned");
+            if (status.is2xxOK()) {
+                // Then check if all partitions are assigned
+                for (auto it = collections.find(name); it != collections.end(); ++it) {
```
jfunston, line 235: I don't think a for loop is correct here

src/k2/cpo/client/Client.h
```diff
@@ -202,54 +220,76 @@ class CPOClient {
...
+                    for (auto& partition : it->second->getAllPartitions()) {
+                        if (partition.astate == dto::AssignmentState::NotAssigned) {
```
jfunston, line 237: Should check if it is Assigned here since it could be in other states like pending.

jfunston (changes requested):

src/k2/cpo/client/Client.h
ccjeff: Ahh my bad - Should be an if check

jfunston (approved):

iafuture (approved):

Issue #295, opened 2022-05-09
Fixes for minimum transaction time


jfunston:
Currently the minimum transaction time is statically coded to 50us in the k23si client.

- TSO should report max error bound to CPO
- CPO should report max error bound among all TSOs to k23si client
- k23si client should use this error bound as the minimum transaction time
- minimum transaction time needs to be implemented for read only transactions as well
- Run YCSB benchmarks, we should see an improvement to the write heavy workloads

ahsank:
@jfunston How TSO knows its error bound to report to CPO? Will CPO get all TSO clock values and compare with CPO clock to find relative drift? Let if know if the process is described in anywhere.

jfunston:
I think to simplify for now, let's skip the steps for getting the error bounds and aggregating from the TSOs. So instead let's add a new configuration variable to the CPO for the minimum transaction time with a default of 20us. This value still needs to be communicated to the client. I think the TSO client should be the component to get it from the CPO as part of its bootstrap ( https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/tso/client/Client.cpp#L62 ). Then the K23SI client should get it from the TSO client.

ahsank:
@jfunston  Few questions

1. Is the error bound value same for all TSOs or TSO services or can be different?
2. I don't see TSO calling CPO at startup, should it also call CPO to get this config?

PR #339, opened 2022-08-16
added k23si get minTransTime from TSO client

ccjeff:
As mentioned in #295, replacing k23si client minimum transaction time to be using the error bound assignment value from CPO.

### DIFF
--- a/src/k2/module/k23si/client/k23si_client.cpp
+++ b/src/k2/module/k23si/client/k23si_client.cpp
@@ -285,19 +285,16 @@ seastar::future<EndResult> K2TxnHandle::end(bool shouldCommit) {
 
             K2LOG_D(log::skvclient, "txn {} end request was accepted", _mtr);
             return _heartbeat_timer.stop().then([this, s=std::move(status)] () {
-                // TODO get min transaction time from TSO client
                 auto time_spent = Clock::now() - _start_time;
                 K2LOG_D(log::skvclient, "time {}, spent: {}", _mtr, time_spent);
-                return _client->getTSOErrorbound()
-                .then([time_spent, s] (auto&& minTransTime) {
-                    if (time_spent < minTransTime) {
-                        auto sleep = minTransTime - time_spent;
-                        return seastar::sleep(sleep).then([s=std::move(s)] () {
-                            return seastar::make_ready_future<EndResult>(EndResult(std::move(s)));
-                        });
-                    }
-                    return seastar::make_ready_future<EndResult>(EndResult(std::move(s)));
-                });
+                auto minTransTime = _client->getTSOErrorbound();
+                if (time_spent < minTransTime) {
+                    auto sleep = minTransTime - time_spent;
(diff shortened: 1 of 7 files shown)


src/k2/module/k23si/client/k23si_client.h
```diff
@@ -135,6 +136,7 @@ class K23SIClient {
     seastar::future<GetSchemaResult> getSchema(const String& collectionName, const String& schemaName, int64_t schemaVersion);
     seastar::future<CreateSchemaResult> createSchema(const String& collectionName, dto::Schema schema);
     seastar::future<CreateQueryResult> createQuery(const String& collectionName, const String& schemaName);
+    seastar::future<Duration> getTSOErrorbound();
```
jfunston, line 139: Let's make this and the TSO call just normal functions, no need for a future here since we know the error bound will be available if we have TSOs to talk to.

src/k2/module/k23si/client/k23si_client.cpp
```diff
@@ -288,14 +288,16 @@ seastar::future<EndResult> K2TxnHandle::end(bool shouldCommit) {
                 // TODO get min transaction time from TSO client
```
jfunston, line 288: Can remove this TODO now

jfunston (changes requested):
You also need to add the minimum wait for read-only transactions (the "if _write_ranges.empty() clause above in the end() call.

src/k2/module/k23si/client/k23si_client.cpp
```diff
@@ -241,9 +241,16 @@ seastar::future<EndResult> K2TxnHandle::end(bool shouldCommit) {
...
+        auto minTransTime = _client->getTSOErrorbound();
+        if (time_spent < minTransTime) {
+            auto sleep = minTransTime - time_spent;
+            return seastar::sleep(sleep).then([] () {
+                return seastar::make_ready_future<EndResult>(EndResult(Statuses::S200_OK("default end result")));
+            });
+        }
         reporter.report();
-        _client->_txnDuration.add(k2::Clock::now() - _startTime);
+        _client->_txnDuration.add(time_spent);
```
jfunston, line 253: Need to add to the txnDuration and report() in the sleep path above too.

jfunston (changes requested):

src/k2/module/k23si/client/k23si_client.cpp
ccjeff: Yeah, I tried to add the report and Prometheus add before but it gives assertion errors in both tpcc test and ycsb test. Will look into it tomorrow.

ccjeff: hmm it seems like for read only transaction that is required to sleep for the minimum transaction time, the `Clock::now()-_start_time` gives a value < 0 and fails the assertion in Prometheus.

jfunston (approved):
Looks good, thanks!

PR #343, opened 2022-09-13
Add getters for schema and collectionName in SKVRecordBuilder

jfunston:
This will make a cleaner interface for usage in chogori-opengauss

### DIFF
--- a/src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
+++ b/src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
@@ -54,6 +54,14 @@ void SKVRecordBuilder::serializeNull() {
     ++_record.fieldCursor;
 }
 
+std::shared_ptr<Schema> SKVRecordBuilder::getSchema() {
+    return _record.schema;
+}
+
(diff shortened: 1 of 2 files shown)


src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
```diff
@@ -54,6 +54,14 @@ void SKVRecordBuilder::serializeNull() {
     ++_record.fieldCursor;
 }
 
+std::shared_ptr<Schema> SKVRecordBuilder::getSchema() {
+    return _record.schema;
+}
+
+String SKVRecordBuilder::getCollectionName() {
```
iafuture, line 61: const& ?

iafuture (changes requested):

src/skvhttpclient/src/skvhttp/dto/SKVRecord.cpp
jfunston: good idea, added it

iafuture (approved):

PR #346, opened 2022-09-20
Change decimal library

dlifw:
Use the str method of the cpp_dec_float class to serialize the object to an std string.
The string is wrapped inside a Binary for write. 
When read, a Binary is read in first. Then, a string is obtained.
The cpp_dec_float class can then be reconstructed from a c string.
(The wrapped string includes the trailing '\0' character for easier handling.)


src/k2/transport/Payload.cpp
```diff
@@ -194,19 +194,21 @@ bool Payload::read(String& value) {
...
-bool Payload::read(std::decimal::decimal64& value) {
-    std::decimal::decimal64::__decfloat64 data;
-    bool success = read((void*)&data, sizeof(data));
+bool Payload::read(boost::multiprecision::cpp_dec_float_50& value) {
+    Binary binData;
+    bool success = read(binData);
     if (!success) return false;
-    value.__setval(data);
+    boost::multiprecision::cpp_dec_float_50 val(binData.get());
+    value = val;
```
iafuture, line 202: combine these in one line to make more efficient: value = boost::multiprecision::cpp_dec_float_50 val(binData.get());

Also, according to the boost API, this can throw. You need to catch std::exception and convert to false return value

src/k2/transport/Payload.cpp
```diff
@@ -194,19 +194,21 @@ bool Payload::read(String& value) {
...
-bool Payload::read(std::decimal::decimal128& value) {
-    std::decimal::decimal128::__decfloat128 data;
-    bool success = read((void*)&data, sizeof(data));
+bool Payload::read(boost::multiprecision::cpp_dec_float_100& value) {
+    Binary binData;
+    bool success = read(binData);
     if (!success) return false;
-    value.__setval(data);
+    boost::multiprecision::cpp_dec_float_100 val(binData.get());
+    value = val;
```
iafuture, line 211: same as above

src/k2/transport/Payload.h
```diff
@@ -531,16 +532,32 @@ class Payload {
         return size + sizeof(size);
     }
 
-    // for type: std::decimal::decimal64
+    // for type: boost::multiprecision::cpp_dec_float_50
     template <typename T>
-    std::enable_if_t<std::is_same_v<T, std::decimal::decimal64>, size_t> getSerializedSizeOf() {
-        return sizeof(std::decimal::decimal64::__decfloat64);
+    std::enable_if_t<std::is_same_v<T, boost::multiprecision::cpp_dec_float_50>, size_t> getSerializedSizeOf() {
```
iafuture, line 537: you can now combine these two methods since the api is the same for both dec50 and dec100. Use the template switch to check for either dev50 || dec100

src/k2/transport/Payload.h
```diff
@@ -531,16 +532,32 @@ class Payload {
...
+        auto curPos = getCurrentPosition();
+        size_t size = 0;
+        if (!read(size)) {
+            K2LOG_E(log::tx, "failed to read payload size");
```
iafuture, line 541: log message should say "decimal size" not "payload size"

src/logging/src/k2/logging/FormattingUtils.h
```diff
@@ -156,15 +156,13 @@ constexpr auto type_name() {
...
 // Provide formatting for decimals
 #ifdef _GLIBCXX_USE_DECIMAL_FLOAT
-#include <decimal/decimal>
+#include <boost/multiprecision/cpp_dec_float.hpp>
 namespace std {
-inline ostream& operator<<(ostream& os, const decimal::decimal64& d) {
-    decimal::decimal64::__decfloat64 data = const_cast<decimal::decimal64&>(d).__getval();
-    return os << (double)data;
+inline ostream& operator<<(ostream& os, const boost::multiprecision::cpp_dec_float_50& d) {
+    return os << d.str();
```
iafuture, line 162: do you need to convert to str first? or can you just do "os << d;" ?

src/logging/src/k2/logging/FormattingUtils.h
```diff
@@ -156,15 +156,13 @@ constexpr auto type_name() {
...
 }
-inline ostream& operator<<(ostream& os, const decimal::decimal128& d) {
-    decimal::decimal128::__decfloat128 data = const_cast<decimal::decimal128&>(d).__getval();
-    return os << (double)data;
+inline ostream& operator<<(ostream& os, const boost::multiprecision::cpp_dec_float_100& d) {
+    return os << d.str();
```
iafuture, line 165: same as above

src/skvhttpclient/src/skvhttp/dto/FieldTypes.h
```diff
@@ -104,14 +104,14 @@ bool isNan(const T& field){
...
+    if constexpr (std::is_same_v<T, DecimalD50>)  { // handle NaN decimal
+        if (field.backend().isnan()) {
             return true;
         }
     }
 
-    if constexpr (std::is_same_v<T, Decimal128> )  { // handle NaN decimal
-        if (std::isnan(Decimal128_to_float(field))) {
+    if constexpr (std::is_same_v<T, DecimalD100> )  { // handle NaN decimal
+        if (field.backend().isnan()) {
```
iafuture, line 114: combine these with ||

src/skvhttpclient/src/skvhttp/mpack/MPackSerialization.h
```diff
@@ -288,37 +288,29 @@ class MPackNodeReader {
...
+        if (!_readFromNode(binData)) {
             return false;
         }
-        if (sizeof(Decimal64::__decfloat64) != sz) {
-            return false;
-        }
-
-        value.__setval(*((Decimal64::__decfloat64*)data));
+        DecimalD50 val = DecimalD50(binData.data());
+        value = val;
```
iafuture, line 300: skip one copy by assigning directly: value = DecimalD50(binData.data());

src/skvhttpclient/src/skvhttp/mpack/MPackSerialization.h
```diff
@@ -288,37 +288,29 @@ class MPackNodeReader {
...
+        if (!_readFromNode(binData)) {
             return false;
         }
-        if (sizeof(Decimal128::__decfloat128) != sz) {
-            return false;
-        }
-
-        value.__setval(*((Decimal128::__decfloat128*)data));
+        DecimalD100 val = DecimalD100(binData.data());
+        value = val;
```
iafuture, line 313: combine into single line

test/transport/PayloadTest.cpp
```diff
@@ -373,7 +373,8 @@ SCENARIO("test getSerializedSizeOf method") {
     Payload src(Payload::DefaultAllocator(32));
```
iafuture, line 373: Maybe just add one more test for bad data. I'd serialize a bad string (e.g. "abcd") and then assert that read should return false when you attempt to read a Decimal50/100

iafuture (changes requested):

jfunston (changes requested):

src/skvhttpclient/src/skvhttp/mpack/MPackSerialization.h
```diff
@@ -487,15 +479,15 @@ class MPackNodeWriter {
...
+        auto shp = std::make_shared<String>(value.str());
+        write(Binary(shp->data(), shp->size() + 1, [shp]() mutable {}));
     }
-    void write(const Decimal128& value) {
-        K2LOG_V(log::mpack, "writing decimal128 type {}", value);
-        Decimal128::__decfloat128 data = const_cast<Decimal128&>(value).__getval();
-        mpack_write_bin(&_writer, (const char*)&data, sizeof(Decimal128::__decfloat128));
+    void write(const DecimalD100& value) {
+        K2LOG_V(log::mpack, "writing decimald100 type {}", value);
+        auto shp = std::make_shared<String>(value.str());
```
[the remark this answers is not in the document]
dlifw: I explored these import and export methods. They seem to be only available for other types, but not available for the cpp_dec_float types. 
The internal representation of the cpp_dec_float is an array of digits, a sign, and an exponent. It hides these internals though.
Also, the str() method is quite smart. If a number is "10,000,000,000.000", the str() method will not return a long string; instead, it returns "1e10".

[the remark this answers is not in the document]
jfunston: I'm not too worried about the number of characters, just the computational cost. Do we have access to the internal representation so we can just memcpy it?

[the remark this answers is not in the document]
dlifw: > 

For the shared pointer, I believe it can help me get a non-owning Binary, and thus avoid one copy operation. Correct me if I am wrong.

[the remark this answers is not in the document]
dlifw: The cpp_dec_float types have fixed sizes and only use stack memory, according to the information on the boost website.
I created a new branch that uses the memcpy to deal with the cpp_dec_float types. 
https://github.com/futurewei-cloud/chogori-platform/compare/master...change-decimal-library-memcpy
Initially, I encountered various issues. After type casting at various places, it finally works. I am not sure whether this is OK.

src/k2/transport/Payload.cpp
dlifw: Changes have been pushed here:
https://github.com/futurewei-cloud/chogori-platform/compare/master...change-decimal-library

src/skvhttpclient/src/skvhttp/common/StreamBuf.h
```diff
@@ -0,0 +1,94 @@
...
+        return ch;
+    }
+
+    ~WriteStreamBuf() {
+        Binary bin(_data, _sz, [](){});
+        _writer.write(bin);
+    }
+private:
+    WriterT& _writer;
+    char _data[bufSize];
```
jfunston, line 61: Can we make this a vector that we call "resize(bufSize)" on and resize if needed? As it is we have a very complicated serialization scheme that still relies on some underlying assumptions about the decimal types that we don't have control over.

jfunston (changes requested):

src/skvhttpclient/src/skvhttp/common/StreamBuf.h
dlifw: Thanks for the review! I considered using vector before, but Ivan told me to not to use it for efficiency reasons.
My concern is that vector may not copy bytes in bulk. If it always read/write individual bytes, the performance may be poor.
Btw, we are not assuming that the serialized size is exactly bufSize bytes. We are just allocating enough bytes for the serialized size. Could we just use a larger number for bufSize to be safe?
FYI:
Sizeof(cpp_dec_float_50) = 56; and the serialized size is 63 bytes.
Sizeof(cpp_dec_float_100) = 80; and the serialized size is 87 bytes.

iafuture: The reason I didn't want a vector is that vector will allocate for each decimal that is being serialized. With this approach here, we are using a stack variable which should be considerably faster (considering the small data size) as it doesn't involve the allocator and can be optimized and cached better.

This isn't decimal specific - it is a StreamBuffer implementation which is targeting small data of some consistent size. The assumption i guess is that you should use this class only for serializing data types of small and very consistent size.

For safety, the code is throwing out_of_range exceptions when there is an attempt to store more data than buffer size. I don't think any additional safety is needed.

src/k2/transport/Payload.cpp
```diff
@@ -306,14 +317,18 @@ void Payload::write(const String& value) {
     write(value.data(), size);
 }
 
-void Payload::write(const std::decimal::decimal64& value) {
-    std::decimal::decimal64::__decfloat64 data = const_cast<std::decimal::decimal64&>(value).__getval();
-    write((const void*)&data, sizeof(data));
+void Payload::write(const boost::multiprecision::cpp_dec_float_50& value) {
+    PayloadWriter writer(*this);
+    skv::http::WriteStreamBuf<PayloadWriter, 80> writesb(writer);
```
jfunston, line 322: Lets add a static assert here and for dec100 below that the sizeof for them is what you have observed. That way we get a compile time error instead of a runtime error if they change the implementation to something we don't expect. Then you can keep the WriteStreamBuf implementation how you have it now. https://en.cppreference.com/w/cpp/language/static_assert

jfunston (changes requested):

dlifw:
Merged another branch for this issue.

PR #347, opened 2022-09-21
Add test script to run platforms servers to send commands remotely

ahsank:
Later test script can be added to this file to verify that opengauss related collections /schema etc are created.

### DIFF
--- a/test/integration/run_test_server.sh
+++ b/test/integration/run_test_server.sh
@@ -1,0 +1,59 @@
+#!/bin/bash
+set -e
+
+topname=$(dirname "$0")
+source ${topname}/common_defs.sh
+cd ${topname}/../..
+
+# start CPO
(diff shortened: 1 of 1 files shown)


test/integration/test_http_client.sh
```diff
@@ -1,5 +1,21 @@
 #!/bin/bash
 set -e
+
+WAITTIME=1
```
jfunston, line 4: I think it is better to just copy this script somewhere if you want to use it for opengauss testing. There is no use for a wait time for normal platform integration tests and for opengauss testing you don't want the tests to ever run. You can use "sleep infinity" then the components will run until you use ctrl-c to clean it up

jfunston (changes requested):

test/integration/test_http_client.sh
ahsank: Added a new script to send commands remotely from opengauss.

PR #348, opened 2022-10-04
Script to run k2 cluster

ahsank:


test/integration/run_k2_cluster.sh
```diff
@@ -0,0 +1,85 @@
...
+# core on which to run the TSO poller thread. Pick 4 if we have that many, or the highest-available otherwise
+#TSO_POLLER_CORE=$(( 5 > $NUMCORES ? $NUMCORES-1 : 4 ))
+# poll on random core
+TSO_POLLER_CORE=-1
+PERSISTENCE=${PERSISTENCE:=tcp+k2rpc://0.0.0.0:12001}
+CPO=${CPO:=tcp+k2rpc://0.0.0.0:9000}
+TSO=${TSO:=tcp+k2rpc://0.0.0.0:13000}
+COMMON_ARGS=${COMMON_ARGS:="--reactor-backend=epoll --enable_tx_checksum true --thread-affinity false"}
+HTTP=${HTTP:=tcp+k2rpc://0.0.0.0:20000}
+CPO_ADDITIONAL_PORT=${CPO_ADDITIONAL_PORT:=9001}
```
iafuture, line 18: This won't work - just remove this from here and the cmdline below. We don't need more than 1 CPO core anyway

test/integration/run_k2_cluster.sh
```diff
@@ -0,0 +1,85 @@
...
+persistence_child_pid=$!
+
+# start tso
+tso ${COMMON_ARGS} -c1 --tcp_endpoints ${TSO} --prometheus_port $(($PROMETHEUS_PORT_START+3)) --tso.clock_poller_cpu=${TSO_POLLER_CORE} &
+tso_child_pid=$!
+
+sleep 1
+
+# start http proxy
+http_proxy ${COMMON_ARGS} -c1 --tcp_endpoints ${HTTP} --memory=1G --cpo ${CPO} --httpproxy_txn_timeout=100h --httpproxy_expiry_timer_interval=50ms --log_level=Info &
```
iafuture, line 40: we should also add the --prometheus_port option here (with port+4 i guess)

iafuture (changes requested):

test/integration/run_k2_cluster.sh
```diff
@@ -0,0 +1,85 @@
...
+
+echo ">>>> Done setting up cluster"
+
+trap finish EXIT
+trap ign INT
+
+if [ -z "$1" ]
+then
+    echo ">>>> Press CTRL+C to exit..."
+    wait ${http_child_pid}
```
jfunston, line 80: I think "sleep infinity" here is better since it will keep the other components up for debugging if the http proxy crashes.

jfunston (changes requested):

test/integration/run_k2_cluster.sh
ahsank: Done

test/integration/run_k2_cluster.sh
ahsank: Done

test/integration/run_k2_cluster.sh
ahsank: Done

jfunston (approved):

iafuture (approved):

PR #354, opened 2022-10-18
HTTP Api to drop collection

ahsank:


src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
```diff
@@ -39,6 +39,15 @@ boost::future<Response<>> Client::createCollection(dto::CollectionMetadata metad
     return _HTTPClient.POST<dto::CollectionCreateRequest>("/api/CreateCollection", std::move(request));
 }
 
+boost::future<Response<>> Client::dropCollection(const String& collectionName) {
+    dto::CollectionDropRequest request{
+        .name = std::move(collectionName)
```
iafuture, line 44: A little surprised this compiles without a warning, but you should not move a const&. Pass by copy instead

src/skvhttpclient/test/ClientIntegrationTest.cpp
```diff
@@ -526,6 +526,21 @@ void testReadWriteConflict() {
   }
 }
 
+void testDropCollection() {
```
iafuture, line 529: Please also add a test to validate that we can create a new collection with the same name after the old one is dropped

iafuture (changes requested):

src/skvhttpclient/src/skvhttp/client/SKVClient.cpp
ahsank: Done

src/skvhttpclient/test/ClientIntegrationTest.cpp
ahsank: Done

jfunston (approved):

PR #357, opened 2022-10-25
use _doGetTSOEndpoints() to set all tso endpoints

dlifw:
Use _doGetTSOEndpoints() in the TSO client to set all the TSO endpoints.

### DIFF
--- a/src/k2/tso/client/Client.cpp
+++ b/src/k2/tso/client/Client.cpp
@@ -73,11 +73,11 @@ seastar::future<> TSOClient::_doGetTSOEndpoints(dto::GetTSOEndpointsRequest &req
         _tsoServerEndpoint = RPC().getTXEndpoint(resp.endpoints[0]);
         for (auto ep : resp.endpoints) {
             _curTSOServiceNodes.push_back(RPC().getTXEndpoint(ep));
+            K2LOG_I(log::tsoclient, "Adding remote TSO endpoint: {}", _curTSOServiceNodes.back()->url);
         }
         std::random_device rd;
         std::mt19937 ranAlg(rd());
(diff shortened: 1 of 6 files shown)


src/k2/tso/client/Client.cpp
```diff
@@ -71,7 +71,20 @@ seastar::future<> TSOClient::_doGetTSOEndpoints(dto::GetTSOEndpointsRequest &req
         }
         _tsoErrorBound = resp.minTransTime;
         _tsoServerEndpoint = RPC().getTXEndpoint(resp.endpoints[0]);
```
jfunston, line 73: _tsoServerEndpoint isn't necessary anymore is it?

src/k2/tso/service/Service.cpp
```diff
@@ -158,12 +158,6 @@ seastar::future<bool> TSOService::_assign(uint64_t tsoID, k2::Duration errBound)
             K2LOG_I(log::tsoserver, "Error bound validation succeeded? {}", isValidErrorBound);
```
iafuture, line 158: Also, remove the _collectWorkerURLs method and usage just above, as well as the _workersURLs member var

iafuture (changes requested):

src/k2/tso/client/Client.cpp
dlifw: Removed.

src/k2/tso/service/Service.cpp
dlifw: Removed.

iafuture (approved):

jfunston (approved):

PR #360, opened 2022-10-27
Fix drop collection taking long time in test

ahsank:
When drop collection, it by default waits 1s to flush remaining transactions that are left. Transactions are not flushed immediately when committed as they are created with sync finalize=false.


johnfangAFW (dismissed):

ahsank:
Also TSO sometime takes 3/4 seconds to initialize clock, causing CPO query fail during this time.

PR #362, opened 2022-11-01
Fix cpo deadline exceeded issue

dlifw:
Add tso_assignment_timeout and assignment_base_backoff for cpo.
    
    The TSO assingment timeout and the base backoff time for the
    ExponentialBackoffStrategy are different concepts.
    Two arguments are added in cpo for users to configure them
    seperately:
    - cpo.tso_assignment_timeout
    - cpo.assignment_base_backoff.
    
    Their default values are both set as 100ms.
    The two arguments are only added in the run_k2_cluster.sh
    integration test explicitly.
    One can use "--cpo.tso_assignment_timeout=100ms"
    and "--cpo.assignment_base_backoff=100ms" to set the values
    explicitly in other tests if needed.

### DIFF
--- a/src/k2/transport/RetryStrategy.cpp
+++ b/src/k2/transport/RetryStrategy.cpp
@@ -41,7 +41,7 @@ ExponentialBackoffStrategy& ExponentialBackoffStrategy::withRate(double rate) {
 
 // Set the desired starting value
 ExponentialBackoffStrategy& ExponentialBackoffStrategy::withBaseBackoffTime(Duration baseBackoffTime) {
-    K2LOG_D(log::tx, "startTimeout: {}ms", k2::msec(baseBackoffTime).count());
+    K2LOG_D(log::tx, "base backoff time: {}", baseBackoffTime);
     _currentBackoffTime = baseBackoffTime;
     return *this;
 }
--- a/src/k2/transport/RetryStrategy.cpp
+++ b/src/k2/transport/RetryStrategy.cpp
@@ -40,9 +40,9 @@ ExponentialBackoffStrategy& ExponentialBackoffStrategy::withRate(double rate) {
 }
 
 // Set the desired starting value
-ExponentialBackoffStrategy& ExponentialBackoffStrategy::withStartTimeout(Duration startTimeout) {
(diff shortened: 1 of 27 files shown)


src/k2/transport/RetryStrategy.cpp
```diff
@@ -40,9 +40,9 @@ ExponentialBackoffStrategy& ExponentialBackoffStrategy::withRate(double rate) {
 }
 
 // Set the desired starting value
-ExponentialBackoffStrategy& ExponentialBackoffStrategy::withStartTimeout(Duration startTimeout) {
-    K2LOG_D(log::tx, "startTimeout: {}ms", k2::msec(startTimeout).count());
-    _currentTimeout = startTimeout;
+ExponentialBackoffStrategy& ExponentialBackoffStrategy::withBaseBackoffTime(Duration baseBackoffTime) {
+    K2LOG_D(log::tx, "startTimeout: {}ms", k2::msec(baseBackoffTime).count());
```
iafuture, line 44: please change this line to just 
```suggestion
    K2LOG_D(log::tx, "startTimeout: {}", baseBackoffTime);
```

src/k2/transport/RetryStrategy.h
```diff
@@ -100,7 +100,7 @@ class ExponentialBackoffStrategy {
                         K2LOG_D(log::txretry, "round {} ended with success={}", _try, _success);
                         if (!_success) {
                             _try++;
-                            _currentTimeout = Duration((uint64_t)(nsec(_currentTimeout).count() * _rate));
+                            _currentBackoffTime = Duration((uint64_t)(nsec(_currentBackoffTime).count() * _rate));
```
iafuture, line 103: I think can also change to 
```suggestion
                            _currentBackoffTime = _currentBackoffTime * _rate;
```

iafuture (changes requested):

src/k2/transport/RetryStrategy.h
dlifw: This will trigger "no match for operator=" error. 
"_currentBackoffTime *= _rate" works fine.

iafuture (approved):

PR #370, opened 2022-12-09
Destroy mpack structs to fix memory leak

ahsank:


src/skvhttpclient/src/skvhttp/mpack/MPackSerialization.h
```diff
@@ -573,8 +623,8 @@ class MPackWriter {
         return true;
     }
 private:
-    char* _data;
-    size_t _size;
-    mpack_writer_t _writer;
+    char* _data = NULL;
```
iafuture, line 626: use {0} syntax instead of = NULL;
same for _size

src/skvhttpclient/src/skvhttp/mpack/MPackSerialization.h
```diff
@@ -554,6 +575,35 @@ class MPackWriter {
...
+        o._size = 0;
+        _writer = std::move(o._writer);
+    }
+
+    MPackWriter& operator=(MPackWriter&& o) {
+        _data = o._data;
+        _size = o._size;
+        o._data = NULL;
+        o._size = 0;
+        _writer = std::move(o._writer);
```
iafuture, line 603: you need to call mpack_writer_destroy as well as MPACK_FREE here (like in the dtor)

iafuture (changes requested):

src/skvhttpclient/src/skvhttp/mpack/MPackSerialization.h
ahsank: Done

iafuture (approved):

PR #374, opened 2023-01-26
Adding configurator feature to dynamically inject configuration changes

mankan1:
Adding configurator feature to dynamically inject configuration changes, added a server and client.


src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
+/*
+MIT License
+
+Copyright(c) 2020 Futurewei Cloud
```
iafuture, line 4: (C)2023

src/k2/transport/RPCTypes.h
```diff
@@ -38,6 +38,9 @@ typedef uint8_t Verb;
 // Verbs used by K2 internally
 enum InternalVerbs : k2::Verb {
     LIST_ENDPOINTS = 249,  // used to discover the endpoints of a node
+    CONFGURATOR_GET = 250,  // used to configure get.
```
iafuture, line 41: You should rename these verbs to CONFIG_SET/GET/DELETE since they manage the config, not the configurator. Also, this enum has the MAX_VERB sentinel value of 250. You cannot repeat it. Since we already have a value for 249, You should list your verbs back from 249, so:
CONFIG_DELETE=246
CONFIG_GET=247
CONFIG_SET=248

src/k2/configurator/ConfiguratorDTO.h
```diff
@@ -0,0 +1,71 @@
...
+    SOFTWARE.
+*/
+
+#pragma once
+#include <k2/common/Common.h>
+#include <k2/transport/PayloadSerialization.h>
+
+namespace k2 {
+
+struct DELETE_ConfiguratorRequest {
```
iafuture, line 30: Let's make the structs match the verb naming. So instead of DELETE_ConfiguratorRequest it should be ConfigDeleteRequest. Same for the rest

src/k2/configurator/ConfiguratorDTO.h
```diff
@@ -0,0 +1,71 @@
...
+    String key;
+    String value;
+    bool applyToAll;
```
iafuture, line 33: We supply values for all primitive types in structs. This should be applyToAll{false}. Same for the rest

src/k2/configurator/ConfiguratorDTO.h
```diff
@@ -0,0 +1,71 @@
...
+    K2_PAYLOAD_FIELDS(key, value, applyToAll);
+    K2_DEF_FMT(DELETE_ConfiguratorRequest, key, value, applyToAll);
+};
+
+struct DELETE_ConfiguratorResponse {
+    String key;
```
iafuture, line 39: no need to return the key.

src/k2/configurator/ConfiguratorDTO.h
```diff
@@ -0,0 +1,71 @@
...
+struct SET_ConfiguratorRequest {
+    String key;
+    String value;
+    bool applyToAll;
+    K2_PAYLOAD_FIELDS(key, value, applyToAll);
+    K2_DEF_FMT(SET_ConfiguratorRequest, key, value, applyToAll);
+};
+
+struct SET_ConfiguratorResponse {
+    String key;
```
iafuture, line 53: no need to return key

src/k2/configurator/ConfiguratorDTO.h
```diff
@@ -0,0 +1,71 @@
...
+    K2_PAYLOAD_FIELDS(key);
+    K2_DEF_FMT(SET_ConfiguratorResponse, key);
+};
+
+struct GET_ConfiguratorRequest {
+    String key;
```
iafuture, line 59: no need to return key

src/k2/configurator/ConfiguratorDTO.h
```diff
@@ -0,0 +1,71 @@
+/*
+MIT License
+
+Copyright(c) 2020 Futurewei Cloud
```
iafuture, line 4: (C) 2023

src/k2/configurator/Configurator.h
```diff
@@ -0,0 +1,46 @@
+/*
+MIT License
+
+Copyright(c) 2020 Futurewei Cloud
```
iafuture, line 4: (C) 2023

src/k2/configurator/Configurator.h
```diff
@@ -0,0 +1,46 @@
...
+public :
+    Configurator();
+    ~Configurator();
+
+    seastar::future<> gracefulStop();
+    seastar::future<> start();
+
+
+private :
+    ConfigVar<String> _key{"key"};
```
iafuture, line 40: Not sure what these are for. Probably remove as well as the Config header above

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include <k2/common/Common.h>
+#include <k2/logging/Log.h>
+#include <k2/common/Common.h>
+#include <k2/transport/PayloadSerialization.h>
+#include <k2/transport/Status.h>
+
+namespace k2 {
+
+namespace log {
+inline thread_local k2::logging::Logger configurator("k2::configurator_server");
```
iafuture, line 39: I would just call the name "k2::config". This name shows up in the logs so something shorter would be better.

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+Configurator::~Configurator() {
+    K2LOG_I(log::configurator, "dtor");
+}
+
+seastar::future<> Configurator::start() {
+    K2LOG_I(log::configurator, "Registering message handlers");
+
+    RPC().registerRPCObserver<SET_ConfiguratorRequest, SET_ConfiguratorResponse>(InternalVerbs::CONFGURATOR_SET, [this](SET_ConfiguratorRequest&& request) {
+                K2LOG_I(log::configurator, "Received set for key: {}", request.key);
```
iafuture, line 54: log the whole request

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+                    auto& conf = const_cast<config::BPOVarMap&>(Config());
+                    conf.emplace(key, boost::program_options::variable_value(value, ""));
+                    boost::program_options::notify(conf);
+                    SET_ConfiguratorResponse response{.key=std::move(request.key)};
+                    return RPCResponse(Statuses::S201_Created("Configurator set accepted"), std::move(response));
+                }
+            });
+
+    RPC().registerRPCObserver<GET_ConfiguratorRequest, GET_ConfiguratorResponse>(InternalVerbs::CONFGURATOR_GET, [this](GET_ConfiguratorRequest&& request) {
+                K2LOG_I(log::configurator, "Received get for key: {}", request.key);
```
iafuture, line 76: log the whole request

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+                if (iter != Config().end()) {    
+                    response.value = (iter->second).as<String>();
+                } else {
+                    return RPCResponse(Statuses::S404_Not_Found("key not found"), std::move(response));
+                }
+                return RPCResponse(Statuses::S200_OK("get accepted"), std::move(response));
+            });
+
+    RPC().registerRPCObserver<DELETE_ConfiguratorRequest, DELETE_ConfiguratorResponse>(InternalVerbs::CONFGURATOR_DELETE, [this](DELETE_ConfiguratorRequest&& request) {
+                K2LOG_I(log::configurator, "Received delete for key: {}", request.key);
```
iafuture, line 91: log the whole request

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+
+#include "Configurator.h"
+#include "ConfiguratorDTO.h"
+
+
+#include <k2/appbase/AppEssentials.h>
+#include <k2/appbase/Appbase.h>
+#include <k2/common/Common.h>
+#include <k2/logging/Log.h>
+#include <k2/common/Common.h>
```
jfunston, line 32: Common.h included twice

src/k2/configurator/Configurator.cpp
jfunston, line 4: Copyright date here and in other new files should be 2023

src/k2/configurator/Configurator.h
jfunston, line 40: Are these variables used anywhere?

src/k2/configurator/ConfiguratorDTO.h
jfunston, line 39: Does the key need to be in the response for any of these?

src/k2/transport/RPCTypes.h
```diff
@@ -38,7 +38,10 @@ typedef uint8_t Verb;
 // Verbs used by K2 internally
 enum InternalVerbs : k2::Verb {
     LIST_ENDPOINTS = 249,  // used to discover the endpoints of a node
-    MAX_VERB = 250,  // something we can use to prevent override of internal verbs.
+    CONFGURATOR_GET = 250,  // used to configure get.
+    CONFGURATOR_SET = 251,  // used to configure set.
+    CONFGURATOR_DELETE = 252,  // used to configure clear messages.
```
jfunston, line 43: I'm not sure this is where we want these defined, maybe Ivan has a suggestion.

test/configurator/ConfiguratorTest.cpp
```diff
@@ -0,0 +1,91 @@
...
+}
+
+class ConfiguratorTest {
+    public:
+        seastar::future<> start() {
+
+            K2LOG_I(log::configuratortest, "Registering message handlers");
+            auto ep = RPC().getServerEndpoint(TCPRPCProtocol::proto);
+
+            _singleTimer.setCallback([ep, this] {
```
jfunston, line 21: I don't think you need a timer for this. You can assign the result from the make_ready_future call below to a member variable

test/integration/test_configurator.sh
```diff
@@ -0,0 +1,51 @@
...
+# start persistence
+./build/src/k2/cmd/persistence/persistence ${COMMON_ARGS} -c1 --tcp_endpoints ${PERSISTENCE} --prometheus_port 63002 &
+persistence_child_pid=$!
+
+# start tso
+./build/src/k2/cmd/tso/tso ${COMMON_ARGS} -c1 --tcp_endpoints ${TSO} --prometheus_port 63003 --tso.clock_poller_cpu=${TSO_POLLER_CORE} &
+tso_child_pid=$!
+
+# start CPO
+./build/src/k2/cmd/controlPlaneOracle/cpo_main ${COMMON_ARGS} -c1 --tcp_endpoints ${CPO} --data_dir ${CPODIR} --txn_heartbeat_deadline=10s --prometheus_port 63000 --assignment_timeout=1s --nodepool_endpoints ${EPS[@]} --tso_endpoints ${TSO} --tso_error_bound=100us --persistence_endpoints ${PERSISTENCE} &
```
jfunston, line 20: Your configurator test is self-contained, you don't need to start any of the k2 components

jfunston (changes requested):

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+}
+
+seastar::future<> Configurator::start() {
+    K2LOG_I(log::configurator, "Registering message handlers");
+
+    RPC().registerRPCObserver<SET_ConfiguratorRequest, SET_ConfiguratorResponse>(InternalVerbs::CONFGURATOR_SET, [this](SET_ConfiguratorRequest&& request) {
+                K2LOG_I(log::configurator, "Received set for key: {}", request.key);
+                auto key = request.key;
+                auto value = request.value;
+
```
iafuture, line 57: We probably need to validate the incoming value here before storing it in the config. you should check the boost  map API to see if it does any config-related validation. If not, we should do it ourselves and return the appropriate status back to the RPC caller.

For log-related config, we need to apply the new value(for both set/delete) to the log system. The logic for that is here: https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/appbase/Appbase.cpp#L81 and we may have to refactor it to reuse it.

You can also avoid code duplication like so:
``` c++
return [request=std::move(request)] {
    auto applier = [request] (auto& config) {
        config.emplace(key, boost::program_options::variable_value(value, ""));
        boost::program_options::notify(config);
        return seastar::make_ready_future();
    };
    if (request.applyToAll) {
        return ConfigDist().invoke_on_all(applier);
    }
    return applier();
    }()
    .then([] {
        return RPCResponse(Statuses::S201_Created("ConfigSet accepted"), ConfigSetResponse{});
    });
```

src/k2/configurator/Configurator.cpp
```diff
@@ -0,0 +1,122 @@
...
+                if (request.applyToAll) {
```
iafuture, line 58: On a further thought, I think the validation and application to config should be delegated to the config here: https://github.com/futurewei-cloud/chogori-platform/blob/master/src/k2/config/Config.h

We should probably add the set/delete methods there which can also hide the all/local dispatch

iafuture (changes requested):

src/k2/configurator/Configurator.cpp
mankan1: fixed

